PR #2, opened 2020-06-22
Set up project 

austinteshuba:
Added in pom.xml files with the appengine and servlet dependencies, as well as some boilerplate files for the webapp.

Displays a simple Hello World page when run/deployed. 

Placeholder is present for project ID from GCP, since we do not yet have a GCP project for the capstone.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -0,0 +1,49 @@
...
+    <build>
+        <plugins>
+            <!-- Provides `mvn package appengine:run` for local testing
+                 and `mvn package appengine:deploy` for deploying. -->
+            <plugin>
+                <groupId>com.google.cloud.tools</groupId>
+                <artifactId>appengine-maven-plugin</artifactId>
+                <version>2.2.0</version>
+                <configuration>
+                    <deploy.projectId>INPUT_PROJECT_ID_HERE</deploy.projectId>
```
antmar, line 43: I'd like to find a way of not including the project ID in the code. This is not a properly of the implementation and more just an artifact of the testing environment.

austinteshuba: The documentation says: "The Google Cloud Platform project name to use for this invocation. You must specify a projectId or you can set GCLOUD_CONFIG to use the project set in your gcloud config state."

Once we have our GCP project, we should look into the configuration to avoid this

PR #7, opened 2020-06-24
Prototype/user-interface

tttzach:
Here are some screenshots of the basic responsive user interface

Full screen monitor

Screenshot 2020-06-24 at 6 40 16 PM

Small window

Screenshot 2020-06-24 at 6 41 53 PM

Mobile-sized window

Screenshot 2020-06-24 at 6 42 11 PM


src/main/webapp/index.html
```diff
@@ -1,10 +1,16 @@
 <!DOCTYPE html>
-<html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <title>Google Dashboard</title>
-</head>
+<meta charset="UTF-8">
+<title>Google Dashboard</title>
+<link rel="stylesheet" href="styles.css">
 <body>
```
austinteshuba, line 5: Fun fact - body is also an optional tag. You can remove the opening and closing body tags.

src/main/webapp/styles.css
```diff
@@ -0,0 +1,23 @@
+.flex-container {
+  align-items: center;
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+}
+
+.flex-item {
+  border: 1px solid black;
+  height: 500px;
```
austinteshuba, line 10: Using hardcoded values (numbers, decimals ... words like 'center' are fine) can actually be avoided in CSS. I recommend creating a new file (perhaps `variables.css`) and including your variable values in there. For Example:

```
/*
  In a variable file
*/
::root {
  --variable-name: variable value;
}

/*
  In another file
*/
.a-sample-class-name {
  property: var(--variable-name);
}
```

src/main/webapp/index.html
tttzach: Didn't know that - will do and thanks!

src/main/webapp/styles.css
tttzach: This is a great idea, thanks for the suggestion!

austinteshuba: https://codeburst.io/css-variables-explained-with-5-examples-84adaffaa5bd If you need some more guidance!

tttzach: I managed to change most of the hard-coded values into variables except for `max-width` in `@media`. Seems like we would need a preprocessor for it so it might be best to retain that one hard-coded value. Here's an [article](https://stackoverflow.com/questions/40722882/css-native-variables-not-working-in-media-queries) from StackOverflow that touches on this.

tttzach: `RESOLVED`

src/main/webapp/index.html
tttzach: `RESOLVED`

src/main/webapp/styles.css
```diff
@@ -0,0 +1,23 @@
+.flex-container {
```
gbuenoandrade, line 1: Nit: I think you need to add the header https://github.com/google/new-project#source-code-headers

src/main/webapp/variables.css
```diff
@@ -0,0 +1,6 @@
+:root {
```
gbuenoandrade, line 1: Nit: Header

src/main/webapp/styles.css
tttzach: Sounds good, thanks!

src/main/webapp/variables.css
tttzach: Thanks!

tttzach: `RESOLVED`

src/main/webapp/styles.css
tttzach: `RESOLVED`

PR #6, opened 2020-06-24
Prototype/authentication-with-backend

austinteshuba:
Create OAuth System with Google Sign in that requests access (read only) to Gmail, Calendar, and Tasks. 

Stores three cookies: clientId (which is the OAuth Client ID to be set in the Google Cloud Platform), idToken (which is the idToken returned from Google sign in), and accessToken (the token that must be sent with all HTTP requests to the Gmail, Calendar, and Tasks REST APIs to prove that the user authorized these actions. (See screenshot)
Screenshot 2020-06-24 at 4 33 51 PM

Utility classes in Java are created to 1) verify the legitimacy of a passed userToken (NOTE: this cannot be tested as of now. #5 ), 2) Get a cookie from a request and 3) generate an Authorization header (in the form of "Bearer <bearer_token_here>").

Also not including links to issues in TODO comments just yet. Since we are changing the repo name soon, the links will break soon

PR #17, opened 2020-06-29
Prototype/utility functions

austinteshuba:
Adds some utility functions to help with handling cookies, like adding, finding, and deleting cookies. 

This is required for handling authentication state, which is the next PR. 

Depends on #6


src/main/webapp/utilities.js
```diff
@@ -0,0 +1,74 @@
...
+function addCookie(cookieName, cookieValue, expiryUtcTime) {
+  document.cookie = `${cookieName}=${cookieValue}; expires=${expiryUtcTime};`;
+}
+
+/**
+ * Deletes cookie from document
+ * @param {string} cookieName name of cookie to be deleted
+ */
+function deleteCookie(cookieName) {
+  document.cookie = `${cookieName}=; expires=Thu, 01 Jan 1970 00:00:01 GMT;';`;
```
tttzach, line 73: Maybe a comment to say why this date? i.e. `// set expiry date to before present time` or set a constant to store this date.

src/main/webapp/utilities.js
```diff
@@ -0,0 +1,74 @@
...
+  // If cookie is not found, or the name is duplicated, return null
+  if (matchingCookies.length !== 1) {
+    if (matchingCookies > 1) {
+      // This is unexpected behaviour
+      console.log('Duplicate cookie!');
+    }
+    return null;
+  }
+
+  return matchingCookies[0];
```
tttzach, line 46: I think a comment like in line 26 would be helpful here too, to eliminate the ambiguity of the indices like `[0]`. Also in line 31 and 34.

src/main/webapp/auth.js
```diff
@@ -58,12 +58,12 @@ function onSignIn(googleUser) {
   // Set cookie that contains idToken to authenticate the user
   // Will automatically delete when the access token expires
   // or when the browser is closed
-  document.cookie = `idToken=${idToken}; expires=${expiryUtcTime}`;
+  addCookie('idToken', idToken, expiryUtcTime);
```
tttzach, line 61: This looks clean after wrapping it in a function!

src/main/webapp/utilities.js
```diff
@@ -0,0 +1,74 @@
...
+ * @param {string} cookieName the name of the cookie that needs to be found
+ * @return {string|null} either cookie's value or null if the cookie not found
+ */
+function getCookie(cookieName) {
+  // Cookie strings are in the format of
+  // cookieName1=cookieValue1; cookieName2=cookieValue2; ...
+  const allCookies = document.cookie.split('; ');
+
+  // Get the value of all matching cookies
+  const matchingCookies = allCookies.filter((cookie) => {
```
AymarMuhikira, line 30: Where is `cookie` defined for this function? Did you mean `cookieName`?

src/main/webapp/utilities.js
AymarMuhikira: Second this

src/main/webapp/utilities.js
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/**
+ * Checks if a cookie is present without returning its value
+ * @param {string} cookieName the name of the cookie that needs to be found
+ * @return {boolean} true if present, false otherwise
+ */
+function isCookiePresent(cookieName) {
+  return getCookie(cookieName) !== null;
+}
```
AymarMuhikira, line 56: I don't see this function being called yet. What use case should we anticipate for this?

src/main/webapp/utilities.js
austinteshuba: Agreed

src/main/webapp/utilities.js
austinteshuba: Added!

src/main/webapp/auth.js
austinteshuba: Thanks!

src/main/webapp/utilities.js
austinteshuba: Will be used to check if authentication cookies are present. If they aren't, the site will have to prompt user to authenticate

AymarMuhikira: Ok, so this is to take care of cases like refreshing?

austinteshuba: This function itself won't handle the no authentication case - it will however enable a check which can be used to trigger a reauthentication. 

This is actually used in the next PR

gbuenoandrade:
> Adds some utility functions to help with handling cookies, like adding, finding, and deleting cookies.
> 
> This is required for handling authentication state, which is the next PR.

Please add `depends on #<PR number>` to the end your description

src/main/webapp/utilities.js
```diff
@@ -0,0 +1,78 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// A series of utility functions that will be helpful throughout the front-end
+
+/* eslint-disable no-unused-vars */
```
gbuenoandrade, line 17: Was ESLint complaining about unused funcs?

src/main/webapp/utilities.js
```diff
@@ -0,0 +1,78 @@
...
+function addCookie(cookieName, cookieValue, expiryUtcTime) {
+  document.cookie = `${cookieName}=${cookieValue}; expires=${expiryUtcTime};`;
+}
+
+/**
+ * Deletes cookie from document
+ * @param {string} cookieName name of cookie to be deleted
+ */
+function deleteCookie(cookieName) {
+  // Set time to EPOCH time. Will prompt cookie to immediately expire
```
gbuenoandrade, line 76: Irrelevant nit: you are actually setting the cookie to expire 1 second past Epoch hehe

src/main/webapp/auth.js
```diff
@@ -17,7 +17,7 @@
 // TODO: Manage User Sign In State (Issue #13)
 
 /* eslint-disable no-unused-vars */
-/* global gapi */
+/* global gapi, addCookie */
```
gbuenoandrade, line 20: Don't need to worry about this for this PR: this is one of the main reasons why I like the `require` approach: it reduces the need to rely on global values

src/main/webapp/utilities.js
austinteshuba: Yes, due to the `deleteCookie` function which is used in later, dependent PRs.

src/main/webapp/utilities.js
austinteshuba: Haha fair enough - Fixing

src/main/webapp/auth.js
austinteshuba: I created an issue for this #31 .

I tried to use require, but it was throwing errors since it's "unrecognized". Turns out I have to install CommonJS/find a way to use it. 

I agree with you - this is messy

PR #18, opened 2020-06-29
Prototype/auth-state

austinteshuba:
Handles authentication state changes, by showing/hiding the login button and features as needed. 

Signed Out:
Screenshot 2020-06-30 at 10 21 08 AM

Signed In:
Screenshot 2020-06-30 at 10 21 21 AM

Dependent on PR #17

PR #20, opened 2020-06-30
Prototype/authentication with java library

austinteshuba:
Create helpers and utilities to authenticate the Java Client API. 

Dependent on PR #18


src/main/java/com/google/sps/utility/AuthenticationUtility.java
```diff
@@ -91,6 +99,78 @@ public static String generateAuthorizationHeader(HttpServletRequest request) {
...
+      }
+    } catch (GeneralSecurityException | IOException e) {
+      e.printStackTrace();
+      return null;
+    }
+
+    // Return null if accessToken cannot be found
+    Cookie accessTokenCookie = getCookie(request, "accessToken");
+    String accessToken = "";
+    if (accessTokenCookie != null) {
```
tttzach, line 123: This might be a good place to use [the "return early" pattern](https://www.itamarweiss.com/personal/2018/02/28/return-early-pattern.html)

src/main/java/com/google/sps/utility/AuthenticationUtility.java
```diff
@@ -91,6 +99,78 @@ public static String generateAuthorizationHeader(HttpServletRequest request) {
...
+   * Creates a Credential object to work with the Google API Java Client. Will NOT handle verifying
+   * userId prior to creating credential. Be sure to do this before creating the credential.
+   * Consider using the overloaded method if you need to do a second verification
+   *
+   * @param accessToken String representation of the accessToken to authenticate user
+   * @return a Google credential object that can be used to create an API service instance. null if
+   *     accessToken is empty string
+   */
+  public static Credential getGoogleCredential(String accessToken) {
+    if (accessToken.equals("")) {
```
tttzach, line 142: extreme nit: maybe .isEmpty() instead?

pom.xml
```diff
@@ -55,6 +55,11 @@
             <artifactId>mockito-all</artifactId>
             <version>1.9.5</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.apis</groupId>
+            <artifactId>google-api-services-gmail</artifactId>
```
tttzach, line 60: extreme nit: maybe put this in the feature/gmail-utility branch instead

austinteshuba: Agreed - Fixed!

src/main/java/com/google/sps/utility/AuthenticationUtility.java
austinteshuba: Fixed!

src/main/java/com/google/sps/utility/AuthenticationUtility.java
austinteshuba: Good idea - fixed!

src/main/java/com/google/sps/utility/AuthenticationUtility.java
```diff
@@ -91,6 +99,76 @@ public static String generateAuthorizationHeader(HttpServletRequest request) {
...
+    return new UrlFetchTransport();
+  }
+
+  /**
+   * Get JsonFactory instance. Use this method to get JsonFactory instances for consistency purposes
+   *
+   * @return JacksonFactory instance
+   */
+  public static JsonFactory getJsonFactory() {
+    return new JacksonFactory();
```
gbuenoandrade, line 169: Would it make sense to always return the same instance?

src/main/java/com/google/sps/utility/AuthenticationUtility.java
```diff
@@ -91,6 +99,76 @@ public static String generateAuthorizationHeader(HttpServletRequest request) {
...
+  }
+
+  /**
+   * Get HTTPTransport appropriate for App Engine environment
+   * https://googleapis.dev/java/google-http-client/latest/com/google/api/client/extensions/appengine/http/UrlFetchTransport.html
+   *
+   * @return UrlFetchTransport instance
+   */
+  public static HttpTransport getAppEngineTransport() {
+    return new UrlFetchTransport();
```
gbuenoandrade, line 160: Would it make sense to always return the same instance?

src/test/java/AuthenticationUtilityTest.java
```diff
@@ -125,4 +131,32 @@ public void getAuthHeaderDuplicateTokens() {
     String header = AuthenticationUtility.generateAuthorizationHeader(request);
     Assert.assertNull(header);
   }
+
+  @Test
+  public void getJsonFactory() {
+    // Should return an object that implements the JsonFactory interface
+    Assert.assertTrue(AuthenticationUtility.getJsonFactory() instanceof JsonFactory);
```
gbuenoandrade, line 138: I don't think this test is necessary as the compiler is already doing this for you. (Same below.)

src/test/java/AuthenticationUtilityTest.java
```diff
@@ -125,4 +131,32 @@ public void getAuthHeaderDuplicateTokens() {
...
+  public void getAppEngineTransport() {
+    // Should return an object that implements the HttpTransport interface
+    Assert.assertTrue(AuthenticationUtility.getAppEngineTransport() instanceof HttpTransport);
+  }
+
+  @Test
+  public void getValidCredential() {
+    // Should create a valid Google credential object with accessToken stored
+    Credential googleCredential = AuthenticationUtility.getGoogleCredential(stubbedAccessToken);
+    Assert.assertNotNull(googleCredential);
```
gbuenoandrade, line 151: Nit: redundant check

src/main/java/com/google/sps/utility/AuthenticationUtility.java
austinteshuba: For all intents and purposes, this should be totally fine. 

We would only need this for building service instances, and Google client APIs don't care which implementation we use

https://developers.google.com/resources/api-libraries/documentation/gmail/v1/java/latest/com/google/api/services/gmail/Gmail.Builder.html

src/main/java/com/google/sps/utility/AuthenticationUtility.java
austinteshuba: In this case it's the expected behaviour for an app engine app, according to the documentation for a Gmail builder (the documentation is similar for the Tasks and Calendar builder as well)

https://developers.google.com/resources/api-libraries/documentation/gmail/v1/java/latest/com/google/api/services/gmail/Gmail.Builder.html

src/test/java/AuthenticationUtilityTest.java
austinteshuba: Agreed

src/test/java/AuthenticationUtilityTest.java
austinteshuba: Agreed

gbuenoandrade:
Assigning back to Austin to address open comments

src/main/java/com/google/sps/utility/AuthenticationUtility.java
austinteshuba: Fixed!

src/main/java/com/google/sps/utility/AuthenticationUtility.java
austinteshuba: Fixed!

src/main/java/com/google/sps/utility/AuthenticationUtility.java
```diff
@@ -40,6 +40,13 @@
   // Application Name
   public static final String APPLICATION_NAME = "PUT NAME HERE";
 
+  // Get JsonFactory used for building Google API Service Instances
+  public static final JsonFactory JSON_FACTORY = new JacksonFactory();
+
+  // Get HTTPTransport appropriate for App Engine environment
+  // https://googleapis.dev/java/google-http-client/latest/com/google/api/client/extensions/appengine/http/UrlFetchTransport.html
+  public static final HttpTransport HTTP_TRANSPORT = new UrlFetchTransport();
```
gbuenoandrade, line 48: Nit: Just noticed that they both offer `getDefaultInstance()`. I think that is preferred over creating your own instances. If you follow that route, then I think it makes more sense for you to simply call `getDefaultInstance()` where needed instead of defining these constants.

austinteshuba: Huh interesting - thanks for pointing this out!

PR #21, opened 2020-06-30
Feature/gmail utility

austinteshuba:
Adds core functionality for gmail API. Displays user's messageIds on the client for prototype purposes.

Screenshot 2020-07-03 at 3 47 41 PM

Dependent on #20


src/main/java/com/google/sps/model/MessageFormat.java
```diff
@@ -0,0 +1,21 @@
+package com.google.sps.model;
```
tttzach, line 1: nit: add copyright header

src/test/java/GmailUtilityTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Test Gmail Utility functions TODO: Mock the Gmail Service features (Issue #5) */
+@RunWith(JUnit4.class)
+public final class GmailUtilityTest {
+
+  // Units of time for email query string
+  private final String daysUnit = "d";
```
tttzach, line 12: Use `CONSTANT_CASE` for these constants

src/main/java/com/google/sps/utility/GmailUtility.java
```diff
@@ -0,0 +1,154 @@
...
+   *     or null if either one of the arguments are invalid Trailing space added to properties so
+   *     multiple queries can be concatenated
+   */
+  public static String emailAgeQuery(int emailAge, String emailAgeUnits) {
+    // newer_than:#d where # is an integer will specify to only return emails from last # days
+    if (emailAge > 0 && (emailAgeUnits.equals("h") || emailAgeUnits.equals("d"))) {
+      return String.format("newer_than: %d%s ", emailAge, emailAgeUnits);
+    } else if (emailAge == 0 && emailAgeUnits.isEmpty()) {
+      return "";
+    } else {
```
tttzach, line 73: extreme nit: this `else` can be removed to make it [cleaner]((https://www.itamarweiss.com/personal/2018/02/28/return-early-pattern.html))

src/main/java/com/google/sps/model/MessageFormat.java
austinteshuba: Fixed

src/test/java/GmailUtilityTest.java
austinteshuba: Fixed!

src/main/java/com/google/sps/utility/GmailUtility.java
austinteshuba: Fixed!

src/test/java/GmailUtilityTest.java
austinteshuba: Fixed!

src/main/java/com/google/sps/utility/GmailUtility.java
```diff
@@ -0,0 +1,155 @@
...
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Contains logic that handles GET & POST requests to the Gmail API and transforms those responses
+ * into easily usable Java types
+ */
+public class GmailUtility {
+  // Make constructor private so no instances of this class can be made
+  private GmailUtility() {}
```
AymarMuhikira, line 32: Nit: You don't necessarily need to define an empty constructor. One is automatically generated if you don't define it.

src/main/java/com/google/sps/utility/GmailUtility.java
```diff
@@ -0,0 +1,155 @@
...
+   * @param emailAge emails from the last emailAge [emailAgeUnits] will be returned. Set to 0 to
+   *     ignore filter
+   * @param emailAgeUnits "d" for days, "h" for hours, "" for ignore email
+   * @return string to use in Gmail (either client or API) to find emails that match these criteria
+   *     or null if either one of the arguments are invalid Trailing space added to properties so
+   *     multiple queries can be concatenated
+   */
+  public static String emailAgeQuery(int emailAge, String emailAgeUnits) {
+    // newer_than:#d where # is an integer will specify to only return emails from last # days
+    if (emailAge > 0 && (emailAgeUnits.equals("h") || emailAgeUnits.equals("d"))) {
```
AymarMuhikira, line 69: I see [here](https://support.google.com/mail/answer/7190?hl=en) they define the units as d, m, and y. Are you sure h works?

src/main/java/com/google/sps/utility/GmailUtility.java
austinteshuba: I want to ensure that no one instantiates this class directly, so an explicitly defined empty constructor is required

src/main/java/com/google/sps/utility/GmailUtility.java
austinteshuba: Strange that it isn't defined there - it does work in Gmail's search bar, which means it would also work for the API (the query selector is the same as using the search bar).

I did however (on another note) realize I've added spaces to the search query where it is not permitted. Fixing that now

src/main/java/com/google/sps/utility/GmailUtility.java
AymarMuhikira: I see

src/main/java/com/google/sps/utility/GmailUtility.java
AymarMuhikira: OK. Maybe add a comment specifying why we don't want to consider m and y units as valid for our application.

austinteshuba: Sure - adding

src/main/java/com/google/sps/utility/GmailUtility.java
```diff
@@ -0,0 +1,158 @@
...
+   * @param gmailService instance of Gmail Service with valid credential
+   * @param query conditions applied to the search to limit which emails are returned. "" for no
+   *     restrictions.
+   * @return List of Message objects that contain a message and thread ID
+   * @throws IOException if an issue occurs with the Gmail service
+   */
+  public static List<Message> listUserMessages(Gmail gmailService, String query)
+      throws IOException {
+    List<Message> messages =
+        gmailService.users().messages().list("me").setQ(query).execute().getMessages();
```
gbuenoandrade, line 131: I wonder if the first item of this list is indeed linked to an email. I have no emails in my test account, but I'm still presented with one ID when I access https://21-dot-step7-2020.uc.r.appspot.com/

src/test/java/GmailUtilityTest.java
```diff
@@ -0,0 +1,135 @@
...
+  private static final String ONE_DAY_QUERY =
+      String.format("newer_than:%d%s ", ONE_UNIT_OF_TIME, DAYS_UNIT);
+  private static final String ONE_HOUR_QUERY =
+      String.format("newer_than:%d%s ", ONE_UNIT_OF_TIME, HOURS_UNIT);
+  private static final String UNREAD_EMAILS_QUERY = "is:unread ";
+  private static final String FROM_EMAIL_QUERY = String.format("from:%s ", SAMPLE_EMAIL);
+
+  @Test
+  public void getQueryStringDays() {
+    // Inputting a query for 1 day should result in a oneDayQuery
```
gbuenoandrade, line 53: FFT: You already defined the constant `ONE_DAY_QUERY` to improve readability. Is the comment necessary?
(here and below)

austinteshuba: Removing, but is it okay to have a test function without a comment? I personally think it's pretty clear what the test is doing, but it might not be as obvious for new users of the code. Is it okay to assume they should be able to figure this out?

src/main/java/com/google/sps/model/MessageFormat.java
```diff
@@ -0,0 +1,34 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package com.google.sps.model;
+
+/**
+ * Encapsulates possible values for the "format" query parameter in the Gmail GET message method
+ * FULL: Returns full email message data METADATA: Returns only email message ID, labels, and email
+ * headers MINIMAL: Returns only email message ID and labels; does not return the email headers,
+ * body, or payload. RAW: Returns the full email message data with body content in the raw field as
+ * a base64url encoded string;
```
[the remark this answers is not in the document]
austinteshuba: I didn't know you could do this and still make the enum publicly accessible - cool! Thanks!

src/main/java/com/google/sps/utility/GmailUtility.java
austinteshuba: Hmm - can you take a look at your test account's gmail account? I haven't added any to mine but I do have an email with the subject line "Finish setting up your new Google Account" and 3 others, leaving 4 entries on the panel.

src/test/java/GmailUtilityTest.java
gbuenoandrade: Yeah, it's pretty common to have tests without comments. The "comment" is usually its name. If you want to take a look at guava project tests: https://github.com/google/guava/tree/master/guava-tests/test/com/google/common. (Naturally, sometimes it is necessary to add comments to clarify something, but it's not mandatory to always have comments.)

austinteshuba: Cool :)

src/main/java/com/google/sps/utility/GmailUtility.java
austinteshuba: Fixed this up! Should handle the null case fine as well.

austinteshuba:
Add custom errors @austinteshuba

src/main/java/com/google/sps/utility/GmailUtility.java
```diff
@@ -0,0 +1,181 @@
...
+   *
+   * @param gmailService instance of Gmail Service with valid credential
+   * @param query conditions applied to the search to limit which emails are returned. "" for no
+   *     restrictions.
+   * @return List of Message objects with ID and thread ID (Empty if no messages present)
+   * @throws IOException if an issue occurs with the Gmail service
+   */
+  public static List<Message> listUserMessages(Gmail gmailService, String query)
+      throws IOException {
+    // Null if no messages present. Convert to empty list for ease
```
gbuenoandrade, line 152: Thanks for the comment! It's not clear that `getMessages()` would return null in that the case, so the comment definitely helps

Issue #26, opened 2020-07-03
Create a servlet utility so common operations (Like converting objects to JSON) is handled cleanly


austinteshuba:

austinteshuba:
Also create a servlet utility so common operations (Like converting objects to JSON) is handled cleanly

austinteshuba:
Also, create an abstract class that will handle the verify credential stuff on doGet / doPost

PR #27, opened 2020-07-03
Feature/tasks-utility

austinteshuba:
Created utilities for working with Google Tasks, including functions to get a Tasks service instance, a user's tasklist, and a user's tasks.

Created servlet and simple javascript function to display the user's tasks on the client for prototyping
Screenshot 2020-07-03 at 5 52 45 PM
"Hey!" is a task in the user's account.

Dependent on PR #21


src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+   * Returns taskNames from the user's Tasks account
+   *
+   * @param request Http request from client. Should contain idToken and accessToken
+   * @param response 403 if user is not authenticated, list of taskNames otherwise
+   * @throws IOException if an issue arises while processing the request
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
tttzach, line 44: extreme nit: might be cleaner to break the sections in this function into separate helper functions

src/main/java/com/google/sps/utility/TasksUtility.java
```diff
@@ -0,0 +1,51 @@
+package com.google.sps.utility;
```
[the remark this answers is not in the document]
austinteshuba: Added - and I added to IntelliJ so I stop forgetting

src/main/java/com/google/sps/servlets/TasksServlet.java
austinteshuba: So I do agree with you, but I don't think this is needed here. 

Most of the functionality is going to be erased in this method block in the next day or two anyways - in the MVP, we will never actually extract task names.

The conversion to JSON should probably be a utility - I'll include this in issue #26

AymarMuhikira:
In the PR description:
> Dependent on PR #21
I get that PR #21 is the base branch since the gmail files are included here, but doesn't tasks-utility technically depend more on PR #20 ? Would it not be functional without gmail?

austinteshuba:
> In the PR description:
> 
> > Dependent on PR #21
> 
> I get that PR #21 is the base branch since the gmail files are included here, but doesn't tasks-utility technically depend more on PR #20 ? Would it not be functional without gmail?

You're 100% right - I just indicated this because of the base branch. In reality I probably could have just based this off PR #20 but I wanted to be able to show both in the demo

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -27,7 +27,10 @@
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-/** Serves selected information from the User's Gmail Account */
+/**
+ * Serves selected information from the User's Gmail Account. TODO: Create Servlet Utility to handle
+ * common functions (Issue #26)
```
antmar, line 32: Good job leaving a TODO trail to help find where refactoring needs to be made!

src/main/java/com/google/sps/servlets/TasksServlet.java
antmar: You could define a new abstract class called `AuthenticatedHttpServlet`, which implements `doGet` and defines a new  abstract method `doAuthedGet` which accepts an extra `Credential` parameter.

I agree JSON stuff should be a utility.

src/main/webapp/populateData.js
```diff
@@ -47,3 +47,33 @@ function populateGmail() {
...
+  // Get Container for Tasks content
+  const tasksContainer = document.querySelector('#tasks');
+
+  // Get list of tasks from user's Tasks account
+  // and display the task titles from all task lists on the screen
+  fetch('/tasks')
+      .then((response) => {
+        // If response is a 403, user is not authenticated
+        if (response.status === 403) {
+          throw new Error();
```
antmar, line 63: You may want to define a specific type for authentication errors, and check for this type in the catch below. That way, errors unrelated to authentication can be handled differently (or not at all).

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

src/main/java/com/google/sps/utility/TasksUtility.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.services.tasks.Tasks;
+import com.google.api.services.tasks.model.Task;
+import com.google.api.services.tasks.model.TaskList;
+import java.io.IOException;
+import java.util.List;
+
+public final class TasksUtility {
```
antmar, line 28: Overall, I am concerned this pattern of creating utility methods will become unmanageable. I think `getTasksService` is a good utility function, but `listTaskLists` and `listTasks` can be problematic.
 - The number of utility can become large, if you need to support different options for listing things.
 - These methods are hard to mock in tests, because you have to supply a mock of this complex Tasks service.

The point about testing is the real issue in my opinion.

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -0,0 +1,70 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Serves selected information from the User's Tasks Account. TODO: Create Servlet Utility to handle
+ * common functions (Issue #26)
+ */
+@WebServlet("/tasks")
+public class TasksServlet extends HttpServlet {
```
antmar, line 37: Are you going to add tests for this class?

src/main/java/com/google/sps/servlets/TasksServlet.java
austinteshuba: I like the Abstract class idea - Issue #37

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: Thank you!

src/main/java/com/google/sps/utility/TasksUtility.java
austinteshuba: This will be changed in another PR #39

src/main/java/com/google/sps/servlets/TasksServlet.java
austinteshuba: This will be changed in another PR #39

src/main/webapp/populateData.js
austinteshuba: Fixed! Thanks for the suggestion - I didn't know this was also a Javascript feature

src/main/java/com/google/sps/utility/TasksUtility.java
austinteshuba: This will be handled in the next PR #39

src/main/java/com/google/sps/servlets/TasksServlet.java
austinteshuba: This will be handled in the next PR #39

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: Thanks!

PR #32, opened 2020-07-06
Try accessing Google Secret from cloudbuild.yaml

tttzach:
Just testing


tttzach:
This pull request is no longer needed as it is replaced by #51

PR #35, opened 2020-07-07
Prototype/secret manager with java library

tttzach:
Test


tttzach:
This is redundant because tests are run from GitHub engine so authImplicit will never work as clarified in the standup on 7/7/2020.

PR #40, opened 2020-07-08
Implement JSON utility class

tttzach:
Implemented JSON utility class.


src/main/java/com/google/sps/utility/JsonUtility.java
```diff
@@ -0,0 +1,39 @@
...
+   * Sends data to a HTTP Servlet Response.
+   *
+   * @param response HTTP Servlet Response that data is sent to.
+   * @param object Contains the data that is to be sent.
+   * @throws IOException
+   */
+  public static void sendJson(HttpServletResponse response, Object object) throws IOException {
+    Gson gson = new Gson();
+    String json = gson.toJson(object);
+    response.getWriter().println(json);
```
austinteshuba, line 37: Just curious - does this work? 

You are writing to the response object (which is magically handled within the servlet), but I don't think that "magic" would work here since you aren't returning the response object (you're essentially just writing to a response object and then disposing of it at the end of the function).

I think it would suffice to return the JSON string

src/main/java/com/google/sps/utility/JsonUtility.java
```diff
@@ -0,0 +1,39 @@
...
+  private JsonUtility() {}
+
+  /**
+   * Sends data to a HTTP Servlet Response.
+   *
+   * @param response HTTP Servlet Response that data is sent to.
+   * @param object Contains the data that is to be sent.
+   * @throws IOException
+   */
+  public static void sendJson(HttpServletResponse response, Object object) throws IOException {
```
austinteshuba, line 34: You should write some tests to make sure that valid JSON is enforced for this method.

src/main/java/com/google/sps/utility/JsonUtility.java
austinteshuba: As an add-on, this method does not enforce that the response is set to be of type JSON. This could lead to errors if misused

tttzach: At least in all the cases where I have used it in my starter projects, writing to a response object and disposing it did exactly what I wanted it to do but feel free to let me know if there's a reason why returning a JSON string would be ideal. Also, you having to question whether this works is on me for not writing tests before requesting review - sorry! I'll add one soon. I also agree that the content type should be set.

src/main/java/com/google/sps/utility/JsonUtility.java
tttzach: Agreed, I'll be sure to do so!

src/main/java/com/google/sps/utility/JsonUtility.java
tttzach: I agree, thanks for the suggestion!

src/test/java/JsonUtilityTest.java
```diff
@@ -0,0 +1,100 @@
...
+import javax.servlet.http.HttpServletResponse;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class JsonUtilityTest {
```
AymarMuhikira, line 36: Consider adding Javadoc description of the class

src/main/java/com/google/sps/utility/JsonUtility.java
```diff
@@ -0,0 +1,40 @@
...
+// limitations under the License.
+
+package com.google.sps.utility;
+
+import com.google.gson.Gson;
+import java.io.IOException;
+import javax.servlet.http.HttpServletResponse;
+
+/** Utility class for JSON operations. */
+public final class JsonUtility {
```
austinteshuba, line 22: Don't fix this here, but I just made a Servlet utility in PR #55 . This would be a great addition - once both PRs are approved we should refactor this into one utility

src/test/java/JsonUtilityTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+  @Test
+  public void sendStringToResponse() throws IOException {
+    // String should be sent successfully and in the expected format
+    JsonUtility.sendJson(response, STRING_OBJECT);
+
+    String expected = gson.toJson(STRING_OBJECT);
+    String actual = stringWriter.toString();
+
+    // "\n" because println is used in JsonUtility.json
```
austinteshuba, line 73: It might be better to not enforce the newline character and instead use `Assert.assertTrue(actual.contains(expected))`. That way it doesn't matter if there is a trailing newline or not, since the inclusion/omission of that character is meaningless to the client anyways.

src/test/java/JsonUtilityTest.java
```diff
@@ -0,0 +1,100 @@
...
+  private final List<Integer> LIST_OBJECT = Arrays.asList(1, 3, 5, 9, 7);
+  private final Map<String, Integer> MAP_OBJECT =
+      new HashMap<String, Integer>() {
+        {
+          put("a", 1);
+          put("c", 3);
+        }
+      };
+
+  static HttpServletResponse response = mock(HttpServletResponse.class);
```
austinteshuba, line 48: Nit: Any particular reason why these aren't private? Doesn't really matter, but might as well stay consistent

src/test/java/JsonUtilityTest.java
```diff
@@ -0,0 +1,100 @@
...
+    Assert.assertEquals(expected + "\n", actual);
+  }
+
+  @Test
+  public void sendListToResponse() throws IOException {
+    // List should be sent successfully and in the expected format
+    JsonUtility.sendJson(response, LIST_OBJECT);
+
+    String expected = gson.toJson(LIST_OBJECT);
+    String actual = stringWriter.toString();
```
austinteshuba, line 83: You should consider using stringWriter.flush() before retrieving the string value, in case if some of the text has not been written and is still in buffer. [Stack Overflow](https://stackoverflow.com/questions/2340106/what-is-the-purpose-of-flush-in-java-streams#:~:text=When%20you%20write%20data%20to,file%2Fbuffered%20writer's%20close()%20.)

src/main/java/com/google/sps/utility/JsonUtility.java
tttzach: Ok got it!

src/test/java/JsonUtilityTest.java
tttzach: That's a great idea, thanks!

src/test/java/JsonUtilityTest.java
tttzach: Nope, I missed it. Thanks for pointing it out!

src/test/java/JsonUtilityTest.java
tttzach: This is a great find, thank you!

src/test/java/JsonUtilityTest.java
tttzach: That's true, thanks Aymar!

src/main/java/com/google/sps/utility/JsonUtility.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/utility/JsonUtility.java
tttzach: `RESOLVED`

src/test/java/JsonUtilityTest.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/utility/JsonUtility.java
tttzach: `RESOLVED`

src/test/java/JsonUtilityTest.java
tttzach: `RESOLVED`

src/test/java/JsonUtilityTest.java
tttzach: `RESOLVED`

src/test/java/JsonUtilityTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+  @Test
+  public void sendStringToResponse() throws IOException {
+    // String should be sent successfully and in the expected format
+    JsonUtility.sendJson(response, STRING_OBJECT);
+
+    String expected = gson.toJson(STRING_OBJECT);
+    String actual = stringWriter.toString();
+
+    Assert.assertTrue(actual.contains(expected));
```
antmar, line 81: Another (better) way to test these methods is to decode the `actual` value and check that the decoded value matches the value that you put in.

It's better because you don't really care that the output is exactly what you expect, as long as the value can be parsed correctly and contains the data that you want. There are some variations in JSON, like whitespace, that you don't really care about.

src/test/java/JsonUtilityTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Test Json Utility methods */
+@RunWith(JUnit4.class)
+public class JsonUtilityTest {
```
antmar, line 38: Are there any failure cases you need to test? What if you pass in a `null` value? What if the value is not a supported type, like some custom Java class?

src/main/java/com/google/sps/utility/JsonUtility.java
```diff
@@ -0,0 +1,40 @@
...
+  private JsonUtility() {}
+
+  /**
+   * Sends data to a HTTP Servlet Response.
+   *
+   * @param response HTTP Servlet Response that data is sent to.
+   * @param object Contains the data that is to be sent.
+   * @throws IOException
+   */
+  public static void sendJson(HttpServletResponse response, Object object) throws IOException {
```
antmar, line 34: Some food for thought: this utility is useful in itself, but it still leaves room for misuse. You could end up calling sendJson twice, which would produce invalid JSON.

Furthermore, the content type can only be changed if no response has been sent yet. If something has already sent a part of the response, the result will be invalid JSON.

Ideally, there should be an abstraction around your servlets that lets you send the response *only once* - either as an error or as a valid JSON object.

src/test/java/JsonUtilityTest.java
tttzach: I agree, this would probably make more sense than checking for every { } , : and whitespace. Thanks!

src/test/java/JsonUtilityTest.java
tttzach: Yes, I should absolutely be testing failure and edge cases. Thanks for pointing this out!

src/main/java/com/google/sps/utility/JsonUtility.java
tttzach: This is an interesting observation. I will definitely look further into this, thanks!

src/test/java/JsonUtilityTest.java
tttzach: I have added two edge cases: `null` and my own custom Java class but I am having trouble finding an unsupported type which causes `sendJson` to fail. Let me know if you have any ideas.

src/main/java/com/google/sps/utility/JsonUtility.java
tttzach: Note: As discussed in today's standup, this would not necessarily be within the scope of this project but would definitely be an important consideration if there were many engineers using the same utility class.

tttzach: `RESOLVED`

src/test/java/JsonUtilityTest.java
tttzach: `RESOLVED`

src/test/java/JsonUtilityTest.java
```diff
@@ -12,94 +12,138 @@
...
     JsonUtility.sendJson(response, MAP_OBJECT);
+    printWriter.flush();
+
+    String actualString = stringWriter.toString();
+    Type type = new TypeToken<Map<String, Integer>>() {}.getType();
+    Map<String, Integer> actual = gson.fromJson(actualString, type);
+
+    Assert.assertEquals(MAP_OBJECT.get("One"), actual.get("One"));
+    Assert.assertEquals(MAP_OBJECT.get("Two"), actual.get("Two"));
+    Assert.assertEquals(MAP_OBJECT.get("Three"), actual.get("Three"));
```
antmar, line 123: I think `Assert.assertEquals(MAP_OBJECT, actual)` should do the right thing. Is there a reason you don't use it?

src/test/java/JsonUtilityTest.java
antmar: Thanks for the new cases. Don't worry about finding an error case - it's really something that Gson should take care of.

src/test/java/JsonUtilityTest.java
```diff
@@ -12,94 +12,138 @@
...
+    // Sends an object of type CustomClass to response as json.
+    // Response should have the same contents as CustomClass object.
+    JsonUtility.sendJson(response, CUSTOM_OBJECT);
+    printWriter.flush();
 
-    String expected = gson.toJson(MAP_OBJECT);
-    String actual = stringWriter.toString();
+    String actualString = stringWriter.toString();
+    Type type = new TypeToken<Map<String, String>>() {}.getType();
+    Map<String, String> actual = gson.fromJson(actualString, type);
```
antmar, line 145: Can you deserialize to the target class:
```
CustomClass actual = gson.fromJson(actualString, CustomClass.class;
```

src/test/java/JsonUtilityTest.java
```diff
@@ -12,94 +12,138 @@
...
+    // Mock servlet response writer
+    response = Mockito.mock(HttpServletResponse.class);
+    stringWriter = new StringWriter();
+    printWriter = new PrintWriter(stringWriter);
+    Mockito.when(response.getWriter()).thenReturn(printWriter);
   }
 
   @After
-  public void clear() {
+  public void tearDown() throws IOException {
```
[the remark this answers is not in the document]
tttzach: If I understand correctly, this means instantiating `stringWriter` and `printWriter` in every test and to me that seems like it would add clutter. Is there another advantage to having `stringWriter` and `printWriter` as instance variables?

src/test/java/JsonUtilityTest.java
tttzach: Awesome, thanks!

src/test/java/JsonUtilityTest.java
tttzach: Thanks for pointing this out! Not sure why I underestimated `assertEquals`.

src/test/java/JsonUtilityTest.java
tttzach: @antmar It seems like I would have to write my own deserializer and then register `CustomClass` as a type adapter in order to run that line of code. Would you recommend this to be done?

antmar: Perhaps I misunderstood the documentation, but it seems like you don't need a custom deserializer: https://github.com/google/gson/blob/master/UserGuide.md#TOC-Object-Examples

Please don't waste too much time on this, however.

src/test/java/JsonUtilityTest.java
antmar: Yeah, `Assert.assertEquals` just uses `object.Equals`, which is correctly defined for `Map`.

src/test/java/JsonUtilityTest.java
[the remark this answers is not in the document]
antmar: JUnit instantiates a new instance of your class for every test, and runs `setUp` before the test method. You don't need to set up the objects in the test method itself - the curreset setup code would still work.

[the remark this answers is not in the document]
tttzach: I didn't know that, thanks for pointing this out and explaining in our standup today!

src/test/java/JsonUtilityTest.java
tttzach: I found the issue by reading the example - it was because CustomClass didn't have a no-args constructor so I simplified CustomClass and deserialized CustomClass to the target class. Thanks for the link Anton!

src/test/java/JsonUtilityTest.java
tttzach: Nice!

PR #49, opened 2020-07-09
Refactor existing style to conform with 7:1 principle of file hierarchy

austinteshuba:
No functional differences to style here - simply renaming of existing files. These files will join a new hierarchy as defined by the 7:1 SASS Principle. 

As the name suggests, this is usually used for SASS (A CSS Pre-processor), but the benefits still persist for CSS - this makes it clear where specific style information lies as the site style gets more complex.


AymarMuhikira:
> These files will join a new hierarchy as defined by the 7:1 SASS Principle.

Out of curiosity, any link to the 7:1 SASS Principle?

austinteshuba:
> > These files will join a new hierarchy as defined by the 7:1 SASS Principle.
> 
> Out of curiosity, any link to the 7:1 SASS Principle?

I learned it as principle, but I see some people calling it a pattern or architecture. Guess it doesn't really matter in the grand scheme of things - it's just a concept that helps you define how files should be split. Here's a [link](https://www.learnhowtoprogram.com/user-interfaces/building-layouts-preprocessors/7-1-sass-architecture)

PR #51, opened 2020-07-09
Handle fetching keys from Secret Manager

gbuenoandrade:
This allows keys defined via Google Cloud Secret Manager to be accessed in runtime using `ApiKeyProvider.get()`.

To deploy locally, create a JSON file containing the keys in `src/main/resources/API_KEYS.json` respecting the following format:
```
{
  "example1" : "value1",
  "example2" : "value2"
}
```

PR #52, opened 2020-07-09
Integrate fetching keys from Secret Manager

tttzach:
Test


tttzach:
This is a duplicate of #51 since changes have been made directly to #51

PR #54, opened 2020-07-10
Create abstract class to handle authentication verification/errors

austinteshuba:
Create abstract class to handle authentication verification, extract credentials, and return 403 errors if required.


src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -0,0 +1,62 @@
...
+   * @throws IOException if there is an issue processing the request
+   */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    googleCredential = AuthenticationUtility.getGoogleCredential(request);
+    if (googleCredential == null) {
+      response.sendError(403, ERROR_403);
+    }
+  }
+}
```
AymarMuhikira, line 62: In what instance would we want to call this instead of doGet? Since both methods implement the same logic, is there a way to create a method, so as to avoid re-writing the same code?

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -34,8 +32,7 @@
  * common functions (Issue #26)
  */
 @WebServlet("/tasks")
-public class TasksServlet extends HttpServlet {
-
+public class TasksServlet extends AuthenticatedHttpServlet {
   /**
    * Returns taskNames from the user's Tasks account
```
AymarMuhikira, line 37: I know this change was not made in this PR, but is this description exact? The json string in the response contains more than task titles

austinteshuba: As of now (in the prototype version) - that's all it did. During the MVP this will be changed to be a bit more feature rich

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: Good idea - I added a helper

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -44,22 +40,21 @@
...
-      return;
-    }
+    // Load and verify google credential
+    super.doGet(request, response);
 
-    // Get messageIds from Gmail
-    Gmail gmailService = GmailUtility.getGmailService(googleCredential);
-    List<Message> messages = GmailUtility.listUserMessages(gmailService, "");
+    // Credential is null if user is not authenticated.
+    if (googleCredential != null) {
```
tttzach, line 47: Just curious - do you want to return an error of some sort to the client if the credential is null like before? I think it would be nice since you're already checking for it.

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -45,26 +42,25 @@
...
 
-    // Get tasks from Google Tasks
-    Tasks tasksService = TasksUtility.getTasksService(googleCredential);
-    List<TaskList> taskLists = TasksUtility.listTaskLists(tasksService);
-    List<Task> tasks = new ArrayList<>();
-    for (TaskList taskList : taskLists) {
-      tasks.addAll(TasksUtility.listTasks(tasksService, taskList));
-    }
+    // Credential is null if user is not authenticated.
+    if (googleCredential != null) {
```
tttzach, line 49: Same as above - do you want to return an error of some sort to the client if the credential is null like before? I think it would be nice since you're already checking for it.

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -45,26 +42,25 @@
...
+      // Get tasks from Google Tasks
+      Tasks tasksService = TasksUtility.getTasksService(googleCredential);
+      List<TaskList> taskLists = TasksUtility.listTaskLists(tasksService);
+      List<Task> tasks = new ArrayList<>();
+      for (TaskList taskList : taskLists) {
+        tasks.addAll(TasksUtility.listTasks(tasksService, taskList));
+      }
```
tttzach, line 56: Would be nice to wrap this in a helper function!

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -0,0 +1,71 @@
...
+   * authenticated
+   *
+   * @param request Http request sent from client
+   * @param response Http response to be sent back to the client
+   * @throws IOException if there is an issue processing the request
+   */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    loadCredential(request, response);
+  }
```
tttzach, line 55: These look very clean!

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -0,0 +1,71 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** HttpServlet that enforces the verification of user tokens */
+public abstract class AuthenticatedHttpServlet extends HttpServlet {
+  // Error message if user is not authenticated
+  protected static final String ERROR_403 = "Authentication tokens not present / invalid";
+
+  protected Credential googleCredential;
```
antmar, line 29: Instead of storing googleCredential as an instance variable, I suggest creating a new abstract `doGet` that takes a Credential as a parameter. You would call it only if there are valid credentials.

This has two benefits:

1. You won't forget to check if `googleCredential` is valid in your servelet.
2. It's really easy to test a servelet by just calling its `doGet` with a mock credential.

austinteshuba: Done! Really cool - thanks

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: Thanks :)

src/main/java/com/google/sps/servlets/TasksServlet.java
austinteshuba: Agreed - done!

src/main/java/com/google/sps/servlets/TasksServlet.java
austinteshuba: Thanks :)

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: Agreed - done!

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -34,37 +33,62 @@
...
-    // get Google credential object. Ensure it is valid - otherwise return an error to client
-    Credential googleCredential = AuthenticationUtility.getGoogleCredential(request);
-    if (googleCredential == null) {
-      response.sendError(403, AuthenticationUtility.ERROR_403);
-      return;
+  public void doGet(
+      HttpServletRequest request, HttpServletResponse response, Credential googleCredential)
+      throws IOException {
+    // Credential is null if user is not authenticated.
+    if (googleCredential != null) {
```
gbuenoandrade, line 50: Please use the return early pattern

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Handle GET request
+   *
+   * @param request HTTP request from client
+   * @param response Http response to be sent to client
+   * @param googleCredential valid, verified google credential object
+   * @throws IOException is there is an issue processing the request
+   */
+  public abstract void doGet(
```
gbuenoandrade, line 71: Any reason for this method to be public instead of protected?

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Handle GET request
+   *
+   * @param request HTTP request from client
+   * @param response Http response to be sent to client
+   * @param googleCredential valid, verified google credential object
+   * @throws IOException is there is an issue processing the request
+   */
+  public abstract void doPost(
```
gbuenoandrade, line 83: Any reason for this method to be public instead of protected?

src/main/java/com/google/sps/servlets/TasksServlet.java
gbuenoandrade: If `googleCredential` should never be null, please throw when it is so we can quickly notice something is wrong. ([Fail fast principle](https://www.martinfowler.com/ieeeSoftware/failFast.pdf).)

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -18,39 +18,32 @@
...
-    if (googleCredential == null) {
-      response.sendError(403, AuthenticationUtility.ERROR_403);
-      return;
-    }
-
+  public void doGet(
+      HttpServletRequest request, HttpServletResponse response, Credential googleCredential)
+      throws IOException {
     // Get messageIds from Gmail
     Gmail gmailService = GmailUtility.getGmailService(googleCredential);
```
gbuenoandrade, line 48: Same as above, assert that the credential is valid

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -62,4 +55,19 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+   * @param request HTTP request from client
+   * @param response Response which will contain 400 error
+   * @param googleCredential valid google credential object (already verified)
+   * @throws IOException if an issue arises while processing the request
+   */
+  @Override
+  public void doPost(
+      HttpServletRequest request, HttpServletResponse response, Credential googleCredential)
+      throws IOException {
+    response.sendError(400, "Post is not supported");
```
gbuenoandrade, line 71: Maybe throw [UnsupportedOperationException](https://docs.oracle.com/javase/6/docs/api/java/lang/UnsupportedOperationException.html) so as to fail fast? (I don't have a strong opinion in this case, it's just food for thought.)

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Verifies user credentials on GET (sending a 403 error in the case that the user is not properly
+   * authenticated).
+   *
+   * @param request Http request sent from client
+   * @param response Http response to be sent back to the client
+   * @throws IOException if there is an issue processing the request
+   */
+  @Override
+  public final void doGet(HttpServletRequest request, HttpServletResponse response)
```
antmar, line 38: Good use of `final` keyword here - forces derived classes to override your new abstract methods.

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
antmar: The original `doGet` and `doPost` methods were also public, and we want the servlet implementations to have public methods that we can unit-test. I am not sure that Java allows changing the visibility of an override.

src/main/java/com/google/sps/servlets/GmailServlet.java
antmar: Alternatively, put this in the implementation for `doPost` in `AuthenticatedHttpServlet` (it won't be an abstract method any more, but that's ok).

That way, derived classes don't need to implement `doPost` if it is not supported.

I don't have a strong feeling about returning a 400 response vs. throwing an exception. The 400 response seems like a more accurate error. (I assume exceptions would result in a 500 response.)

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: Thank you!

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: Throwing an exception would result in a 500 error, which I would prefer to reserve for actual implementation issues on the server side (this error would just be an illegal call). 

I will include this in the AuthenticatedHttpServlet however as a default implementation.

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: Cool idea - implementing

src/main/java/com/google/sps/servlets/TasksServlet.java
austinteshuba: Oops - I forgot to update this servlet once I made changes to the authenticatedHttpServlet. 

This method is called on the condition that the googleCredential is not null, so this check is redundant. Replacing with assertion

src/main/java/com/google/sps/servlets/TasksServlet.java
austinteshuba: Using assertion to remedy this

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: The original methods are actually protected (They used public in the walkthroughs so I defaulted to that). 

We should chat offline about this - switching to protected is probably better semantically, but has some implications for unit testing

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: We should chat offline about this - switching to protected is probably better semantically, but has some implications for unit testing

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -0,0 +1,107 @@
...
+   * @param request HttpRequest from client
+   * @param response Response to send to client
+   * @throws IOException if an issue occurs processing the response
+   */
+  private Credential loadCredential(HttpServletRequest request, HttpServletResponse response)
+      throws IOException {
+    Credential googleCredential = AuthenticationUtility.getGoogleCredential(request);
+    if (googleCredential == null) {
+      response.sendError(403, ERROR_403);
+    }
```
antmar, line 103: It's a little weird for a method called `loadCredentials` to also send a response in some cases. I would prefer to just inline this logic in each of `doGet` and `doPost`, with an early exit if credentials are `null`.

I think a small amount of code duplication in (what is effectively) a utility class is OK.

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -18,38 +18,34 @@
...
-    Credential googleCredential = AuthenticationUtility.getGoogleCredential(request);
-    if (googleCredential == null) {
-      response.sendError(403, AuthenticationUtility.ERROR_403);
-      return;
-    }
+  public void doGet(
+      HttpServletRequest request, HttpServletResponse response, Credential googleCredential)
+      throws IOException {
+    // Null credentials (i.e. unauthenticated requests) should be handled by parent class
+    assert googleCredential != null;
```
gbuenoandrade, line 48: Nit: Turn the comment into an assertion message, e.g.:
`assert x > 0 : "x must be greater than zero"`

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -19,46 +19,39 @@
...
-    Credential googleCredential = AuthenticationUtility.getGoogleCredential(request);
-    if (googleCredential == null) {
-      response.sendError(403, AuthenticationUtility.ERROR_403);
-      return;
-    }
+  public void doGet(
+      HttpServletRequest request, HttpServletResponse response, Credential googleCredential)
+      throws IOException {
+    // Null credentials (i.e. unauthenticated requests) should be handled by parent class
+    assert googleCredential != null;
```
gbuenoandrade, line 50: Nit: same as above

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: I agree - this might be a case of over-abstraction. I'm going to get rid of the helper.

src/main/java/com/google/sps/servlets/TasksServlet.java
austinteshuba: Fixed

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: Cool - didn't know this existed!

PR #55, opened 2020-07-10
Create Servlet Utility to handle common Servlet operations

austinteshuba:
This is a refactor. Servlet utility will now handle common server operations (getting cookies, creating an authorization header, eventually handling JSON and additional functions). 

Dependent on PR #54


src/main/java/com/google/sps/utility/ServletUtility.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletRequest;
+
+/** Contains methods for common servlet operations */
+public final class ServletUtility {
+  // Application Name
```
tttzach, line 25: nit: I think this comment can be removed, it repeats the variable name

austinteshuba: Agreed - thanks!

austinteshuba:
> LGTM. What is the advantage of using the ServletUtility instead of directly using the Authentication utility?

I'm trying to eliminate AuthenticationUtility. Things related to authentication directly should be handled by the abstract class `AuthenticatedHttpServlet`, and general servlet things (which may even exceed the scope of an "Authentication utility") should be handled by the servlet utility.

src/main/java/com/google/sps/utility/ServletUtility.java
```diff
@@ -0,0 +1,77 @@
...
+    if (cookies.isEmpty()) {
+      System.out.println("Cookie not found");
+      return null;
+    }
+
+    // If more than one cookies are found, it is ambiguous as to which one to return.
+    // This is unexpected - duplicate cookies are usually blocked by the browser (especially
+    // when they are user set).
+    if (cookies.size() > 1) {
+      System.out.println("Duplicate cookie");
```
gbuenoandrade, line 51: Nit: logging the contents of `cookies` could be helpful for debugging

src/test/java/ServletUtilityTest.java
```diff
@@ -0,0 +1,128 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.Mockito;
+
+/** Test Authentication Utility functions */
+@RunWith(JUnit4.class)
+public final class ServletUtilityTest {
+
+  // Will mock an HttpServletRequest to be passed to utilities
+  private static final HttpServletRequest request = Mockito.mock(HttpServletRequest.class);
```
gbuenoandrade, line 29: Please do this inside some setup method that gets called before each test using `@Before`. That way you eliminate the risk of tests interfering with each other.

src/main/java/com/google/sps/utility/AuthenticationUtility.java
```diff
@@ -37,9 +34,6 @@
   public static final String CLIENT_ID =
       "12440562259-mf97tunvqs179cu1bu7s6pg749gdpked.apps.googleusercontent.com";
```
gbuenoandrade, line 35: FFT: this would be really easy to hide using Zach's PR (https://github.com/googleinterns/step7-2020/pull/51)

src/test/java/ServletUtilityTest.java
austinteshuba: Agreed - done!

src/main/java/com/google/sps/utility/AuthenticationUtility.java
austinteshuba: Tracking with issue #72

src/main/java/com/google/sps/utility/ServletUtility.java
austinteshuba: Good idea - implementing!

src/main/java/com/google/sps/utility/ServletUtility.java
```diff
@@ -0,0 +1,81 @@
...
+      for (int i = 0; i < cookies.size(); i++) {
+        System.out.println(String.format("Cookie %d: %s", i + 1, cookies.get(i).getValue()));
```
gbuenoandrade, line 53: Nit 1: `System.out.printf` lets you format the string
Nit 2: `list.forEach` to avoid the explicit loop

src/test/java/ServletUtilityTest.java
```diff
@@ -0,0 +1,133 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.Mockito;
+
+/** Test Authentication Utility functions */
+@RunWith(JUnit4.class)
+public final class ServletUtilityTest {
+
+  private static HttpServletRequest request;
```
gbuenoandrade, line 29: Nit: Given that the tests are non-static, I think it'd be better for this to be non-static as well. JUnit could, in theory, create several instances of `ServletUtilityTest` and interact with them in parallel.

src/main/java/com/google/sps/utility/ServletUtility.java
austinteshuba: Nit 1: Cool! Resolved. 
Nit 2: I did initially want the index, but I think the new log is better and cleaner. Thanks for the suggestion!

src/test/java/ServletUtilityTest.java
austinteshuba: Resolved! Thanks for the suggestion

austinteshuba:
> As a general comment, prefer to use exceptions when there is more than one way in which a method can fail. For example, a cookie may be not found or there may be duplicate cookies - the `null` doesn't tell you what happened. Logging is only really helpful during development - in production applications, there are more sophisticated systems for reporting exceptions.

I'm going to resolve this in the next PR :)

PR #38, opened 2020-07-08
Implement initial version of calendar utility

AymarMuhikira:
-Call function to populate calendar panel with events
-Servlet gets the credential and gets the list of events from utility
-Utility makes API call to get calendar service and events.

PR #57, opened 2020-07-10
Initial implementation of calendar utility

AymarMuhikira:
-Replaces PR #38 
-Call function to populate calendar panel with events
-Servlet gets the credential and gets the list of events from utility
-Utility makes API call to get calendar service and events.


src/main/webapp/populateData.js
```diff
@@ -89,3 +89,35 @@ function populateTasks() {
       });
 }
 
+/**
+ * Populate Calendar container with user's events
+ */
+function populateCalendar() {
```
tttzach, line 95: Suggestion for maybe another PR: `populateGmail()`, `populateTasks()` and `populateCalendar()` look very oddly identical to me. It would be nice if we could refactor and write a helper function to do both of these.

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,51 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** GET function responds JSON string containing events in user's calendar. */
+@WebServlet("/calendar")
+public class CalendarServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
tttzach, line 35: nit: Maybe javadoc comment for this method

antmar: You don't generally *have* to add javadoc for overrides, because you should be implementing exactly the interface described by the original method. However, it doesn't hurt to add more documentation.

src/main/java/com/google/sps/utility/CalendarUtility.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.services.calendar.Calendar;
+import com.google.api.services.calendar.model.Event;
+import java.io.IOException;
+import java.util.List;
+
+/** Utility class to use Calendar API instead of handling everything from the servlet */
+public class CalendarUtility {
```
austinteshuba, line 28: This should be final so it cannot be inherited/subclassed

src/main/java/com/google/sps/utility/CalendarUtility.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  /**
+   * Get the list of events in the user's calendar
+   *
+   * @param calendarService a valid Google Calendar service instance
+   * @return the list of Event from the calendar
+   */
+  public static List<Event> getCalendarEvents(Calendar calendarService) throws IOException {
+    return calendarService.events().list("primary").setPageToken(null).execute().getItems();
```
austinteshuba, line 54: I notice that `calendarService.events().list("primary").execute().getItems();` still works. Is there a benefit to setting page token to null?

src/main/webapp/auth.js
```diff
@@ -116,7 +117,7 @@ function signOut() {
 function renderButton() {
   gapi.signin2.render('google-sign-in-btn', {
     'scope': 'https://www.googleapis.com/auth/gmail.readonly ' +
-        'https://www.googleapis.com/auth/calendar.readonly ' +
+        'https://www.googleapis.com/auth/calendar ' +
```
austinteshuba, line 120: I don't see any post methods built out yet for calendar. Is there any particular reason why this needs to be changed?

src/main/webapp/populateData.js
austinteshuba: Just created an issue #68

austinteshuba:
Also, this does not follow the client abstraction pattern we agreed to offline. However, this can be fixed in another PR.

src/main/java/com/google/sps/servlets/CalendarServlet.java
AymarMuhikira: Good to know. I will add the documentation.

src/main/java/com/google/sps/utility/CalendarUtility.java
AymarMuhikira: Agreed. Fixed it

src/main/java/com/google/sps/utility/CalendarUtility.java
AymarMuhikira: Changed to `calendarService.events().list("primary").execute().getItems();`

src/main/webapp/auth.js
AymarMuhikira: For later use. We did specify in the design doc that we wanted to give `read` and `write` authorization to this API. I don't think there is necessarily a benefit in doing it later than now.

austinteshuba: True - It doesn't really matter too much. We can leave it

PR #59, opened 2020-07-10
Implement Directions utility class

tttzach:
This class was created prior to client abstraction which will be addressed on #46 
This pull request replaces #25 to remove all unrelated commit history


src/main/java/com/google/sps/servlets/DirectionsServlet.java
```diff
@@ -0,0 +1,72 @@
...
+  public static void sendJson(HttpServletResponse response, Object object) throws IOException {
+    Gson gson = new Gson();
+    String json = gson.toJson(object);
+    response.setContentType("application/json");
+    response.getWriter().println(json);
+  }
+
+  /** Get directions using the parameters stored in this class. */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
AymarMuhikira, line 49: Nit: you could add the parameters of the method in the Javadoc description

src/main/java/com/google/sps/utility/DirectionsUtility.java
```diff
@@ -0,0 +1,97 @@
...
+   * @param destination A string representing the destination to get directions to.
+   * @param origin A string representing the origin to get directions from.
+   * @param waypoints A list of string consisting of waypoints to visit between the destination and
+   *     the origin.
+   * @param apiKey A string representing the API key to authenticate a Google Directions API call.
+   * @return A string representing the URI.
+   * @throws URISyntaxException Indicates that the built string could not be parsed as a URI.
+   */
+  public static String getDirectionsUri(
+      String destination, String origin, List<String> waypoints, String apiKey)
```
AymarMuhikira, line 61: Does this method need to be accessible outside of the class?

src/main/webapp/directions.js
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/* exported getDirections */
+
+/**
+ * Gets key information from each leg and displays them on the DOM.
+ */
+function getDirections() {
```
AymarMuhikira, line 20: Consider adding this to the populateData script.

src/test/java/DirectionsUtilityTest.java
```diff
@@ -0,0 +1,73 @@
...
+@RunWith(JUnit4.class)
+public class DirectionsUtilityTest {
+
+  private final String ORIGIN = "University of Waterloo";
+  private final String DESTINATION = "Wilfrid Laurier University";
+  private final List<String> NO_WAYPOINTS = Arrays.asList();
+  private final List<String> ONE_WAYPOINT = Arrays.asList("University of Waterloo Place");
+  private final List<String> MANY_WAYPOINTS =
+      Arrays.asList("University of Waterloo Place", "Google Kitchener");
+  private final String API_KEY = "SAMPLE_API_KEY";
```
AymarMuhikira, line 34: Some other edge cases to consider: what if one of the end points is also a waypoint, what if  a waypoint is duplicated in the many waypoints scenario, is it possible to have an unrecognized address? What are the expected behaviours for these cases? Should we test for those>

tttzach: These are all valid edge cases! All of these should be handled in the step before using this utility which involves the Google Places API (handles all the unrecognized addresses) and creates only valid origin, destination and waypoints parameters. I should definitely note this on the file, thanks!

src/main/webapp/directions.js
tttzach: This is a great idea, thanks!

src/main/java/com/google/sps/utility/DirectionsUtility.java
tttzach: Nope, the only reason this is public was only for testing purposes. Should probably test these within this class. Thanks for pointing it out!

src/main/java/com/google/sps/servlets/DirectionsServlet.java
tttzach: Absolutely, thanks for catching that!

src/main/webapp/directions.js
tttzach: `RESOLVED`

src/test/java/DirectionsUtilityTest.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/DirectionsServlet.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+@WebServlet("/directions")
+public class DirectionsServlet extends HttpServlet {
```
austinteshuba, line 31: Public classes require javadoc. [7.3](https://google.github.io/styleguide/javaguide.html#s7.1-javadoc-formatting)

src/main/java/com/google/sps/servlets/DirectionsServlet.java
```diff
@@ -0,0 +1,78 @@
...
+    if (waypointsString == null) {
+      waypoints = Arrays.asList();
+    } else {
+      waypoints = Arrays.asList(waypointsString.split(";"));
+    }
+    try {
+      JSONObject directions =
+          DirectionsUtility.getDirections(destination, origin, waypoints, apiKey);
+      sendJson(response, directions);
+    } catch (JSONException e) {
```
austinteshuba, line 70: These two catch blocks can be simplified by doing: 

`catch (JSONException | URISyntaxException e) {...}`

src/main/java/com/google/sps/utility/DirectionsUtility.java
austinteshuba: If there could be a use for this outside of the class, you could consider making this public.

src/main/java/com/google/sps/servlets/DirectionsServlet.java
tttzach: Thanks for letting me know, I have incorporated this in the new dependency injection ready PR #63 . Thanks!

src/main/java/com/google/sps/servlets/DirectionsServlet.java
tttzach: I have incorporated this in the new dependency injection ready PR #63 . Thanks!

tttzach:
This PR is now a duplicate of #63 since #63 reimplements this pipeline to the Google Directions API in a structure which allows for dependency injection.

PR #61, opened 2020-07-10
Implement JSON utility class

tttzach:
This pull request replaces #40 to remove all unrelated commit history


tttzach:
#40 has comments I would like to keep

PR #69, opened 2020-07-13
Refactor authentication to use client abstraction pattern

austinteshuba:
Includes unit tests for the ClientIDServlet and AuthenticatedHttpServlet. Updated version of mockito to allow mocking of abstract classes directly. 

Dependent on PR #55


austinteshuba:
@austinteshuba Implement more appropriate exception handling where appropriate

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -26,6 +29,24 @@
   // Error message if user is not authenticated
   protected static final String ERROR_403 = "Authentication tokens not present / invalid";
 
+  private final AuthenticationVerifier authenticationVerifier;
+
+  /** Create AuthenticatedHttpServlet with default implementations of the AuthenticationVerifier */
+  public AuthenticatedHttpServlet() {
+    super();
```
tttzach, line 36: This [link](https://stackoverflow.com/questions/4090834/when-do-i-use-super) might provide some insight to using the `super()` keyword.

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -26,6 +29,24 @@
...
+    authenticationVerifier = new AuthenticationVerifierImpl();
+  }
+
+  /**
+   * Create AuthenticatedHttpServlet with an explicit implementation of the AuthenticationVerifier
+   *
+   * @param authenticationVerifier implementation of the AuthenticationVerifier
+   */
+  public AuthenticatedHttpServlet(AuthenticationVerifier authenticationVerifier) {
+    super();
```
tttzach, line 46: This [link](https://stackoverflow.com/questions/4090834/when-do-i-use-super) might provide some insight to using the super() keyword.

src/test/java/AuthenticatedHttpServletTest.java
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  @After
+  public void clear() {
+    // Dump contents after each test
+    stringWriter.getBuffer().setLength(0);
+  }
+
+  @Test
+  public void getRequestNoTokens() throws IOException {
```
tttzach, line 93: Short comments describing what is handled and expected might be useful here.

src/test/java/AuthenticatedHttpServletTest.java
```diff
@@ -0,0 +1,201 @@
...
+      Mockito.mock(
+          AuthenticatedHttpServlet.class,
+          Mockito.withSettings()
+              .useConstructor(authVerifier)
+              .defaultAnswer(Mockito.CALLS_REAL_METHODS));
+
+  private HttpServletRequest request;
+  private HttpServletResponse response;
+  private StringWriter stringWriter;
+  private PrintWriter printWriter;
```
tttzach, line 49: If I recall correctly, there are static or final keywords for these lines on other PRs. Might be useful to have them standardized.

src/test/java/ClientIDServletTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+/** Test Client ID Servlet to ensure response contains defined OAuth 2.0 client ID */
+@RunWith(JUnit4.class)
+public final class ClientIDServletTest {
+  private static final ClientIDServlet servlet = new ClientIDServlet();
+
+  private HttpServletRequest request;
+  private HttpServletResponse response;
+  private StringWriter stringWriter;
+  private PrintWriter printWriter;
```
tttzach, line 40: If I recall correctly, there are static or final keywords for these lines on other PRs. Might be useful to have them standardized.

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -93,4 +114,52 @@ public void doPost(
...
+
+    Cookie accessTokenCookie = ServletUtility.getCookie(request, "accessToken");
+    if (accessTokenCookie == null) {
+      return null;
+    }
+
+    String accessToken = accessTokenCookie.getValue();
+
+    if (accessToken.isEmpty()) {
+      return null;
```
tttzach, line 154: I see a lot of `return null;` here, might be worth throwing an exception of some sort.

src/test/java/AuthenticatedHttpServletTest.java
```diff
@@ -0,0 +1,201 @@
...
+  public void clear() {
+    // Dump contents after each test
+    stringWriter.getBuffer().setLength(0);
+  }
+
+  @Test
+  public void getRequestNoTokens() throws IOException {
+    when(request.getCookies()).thenReturn(noCookies);
+    servlet.doGet(request, response);
+    verify(response, Mockito.times(1)).sendError(Mockito.eq(403), Mockito.anyString());
```
tttzach, line 96: Just out of curiosity, what does `Mockito.times(1)` do in the context of this line?

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: Interesting - thanks

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: Interesting - thanks

src/test/java/AuthenticatedHttpServletTest.java
austinteshuba: I removed those instances. These are not constants and need to be re-declared prior to each test.

src/test/java/ClientIDServletTest.java
austinteshuba: I removed those instances. These are not constants and need to be re-declared prior to each test.

src/test/java/AuthenticatedHttpServletTest.java
austinteshuba: It verifies that the method was called exactly once :)

src/test/java/AuthenticatedHttpServletTest.java
austinteshuba: I'm hesitant to add comments here, as I think they would explain the what not the why. The name of each method describes what's being tested, and the code is pretty trivial in terms of what's expected. Do you have something specific in mind?

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: Agreed - I added exceptions for exceptional circumstances (the tokens not being present is not exceptional)

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -37,18 +56,24 @@
...
+      if (googleCredential == null) {
+        response.sendError(403, ERROR_403);
+        return;
+      }
+      doGet(request, response, googleCredential);
+    } catch (TokenVerificationError e) {
       response.sendError(403, ERROR_403);
-      return;
+    } catch (GeneralSecurityException | IOException e) {
+      response.sendError(500, e.getMessage());
```
antmar, line 69: The default behavior for uncaught exceptions should be to return an error 500. What's your motivation for explicitly catching these?

Separately, it's not a good idea to return the message of exceptions that you didn't create yourself. You don't know what kind of internal implementation details are in that message.

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -57,13 +82,18 @@ public final void doGet(HttpServletRequest request, HttpServletResponse response
...
+      if (googleCredential == null) {
+        response.sendError(403, ERROR_403);
+        return;
+      }
+      doPost(request, response, googleCredential);
+    } catch (TokenVerificationError e) {
       response.sendError(403, ERROR_403);
-      return;
+    } catch (GeneralSecurityException | IOException e) {
+      response.sendError(500, e.getMessage());
```
antmar, line 95: Same comment about catching generic exceptions as in `doGet`

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -93,4 +123,43 @@ public void doPost(
...
+    }
+    String idToken = idTokenCookie.getValue();
+    String accessToken = accessTokenCookie.getValue();
+
+    try {
+      if (!authenticationVerifier.verifyUserToken(idToken)) {
+        throw new TokenVerificationError(String.format("idToken (value=%s) is invalid!", idToken));
+      }
+    } catch (GeneralSecurityException | IOException e) {
+      e.printStackTrace();
```
antmar, line 153: You shouldn't have to manually print the stack trace. It's probably not useful outside of active development anyways.

Ideally, whatever framework runs these servlets will log the stack and other data for unhandled exceptions.

src/main/java/com/google/sps/model/AuthenticationVerifierImpl.java
```diff
@@ -0,0 +1,44 @@
...
+    // Build a verifier used to ensure the passed user ID is legitimate
+    HttpTransport transport = UrlFetchTransport.getDefaultInstance();
+    JsonFactory factory = JacksonFactory.getDefaultInstance();
+    GoogleIdTokenVerifier verifier =
+        new GoogleIdTokenVerifier.Builder(transport, factory)
+            .setAudience(Collections.singletonList(CLIENT_ID))
+            .build();
+
+    // If the userToken is not null, the identity is verified and vice versa
+    GoogleIdToken userToken = verifier.verify(idToken);
```
antmar, line 40: Out of curiosity, do you need to `GoogleIdToken` to call backend APIs, or is it only used for verification?

src/main/java/com/google/sps/utility/ServletUtility.java
```diff
@@ -48,10 +50,10 @@ public static Cookie getCookie(HttpServletRequest request, String cookieName) {
...
     // when they are user set).
     if (cookies.size() > 1) {
-      System.out.printf("%d duplicate cookies with name %s\n", cookies.size(), cookieName);
-      cookies.forEach(c -> System.out.println(c.getValue()));
-
-      return null;
+      String errorMessage =
+          String.format("%d duplicate cookies with name %s\n", cookies.size(), cookieName);
+      errorMessage += cookies.stream().map(Cookie::getValue).reduce((a, b) -> a + "\n" + b);
+      throw new IOException(errorMessage);
```
antmar, line 56: Is `IOException` really the best type of exception to use? This has nothing to do with disk or network operations.

src/test/java/AuthenticatedHttpServletTest.java
```diff
@@ -0,0 +1,172 @@
...
+  private static final Cookie[] noCookies = new Cookie[] {};
+  private static final Cookie[] noAccessTokenCookies = new Cookie[] {sampleIdTokenCookie};
+  private static final Cookie[] noIdTokenCookies = new Cookie[] {sampleAccessTokenCookie};
+  private static final Cookie[] emptyIdTokenCookies =
+      new Cookie[] {emptyIdTokenCookie, sampleAccessTokenCookie};
+  private static final Cookie[] validCookies =
+      new Cookie[] {sampleIdTokenCookie, sampleAccessTokenCookie};
+
+  @Before
+  public void init() throws IOException {
```
antmar, line 71: Prefer to use `setUp` for the method annotated with `@Before`. It's a standard naming convention in JUnit and related frameworks.

src/test/java/AuthenticatedHttpServletTest.java
```diff
@@ -0,0 +1,172 @@
...
+    response = Mockito.mock(HttpServletResponse.class);
+
+    // Writer used in get/post requests to capture HTTP response values
+    stringWriter = new StringWriter();
+    printWriter = new PrintWriter(stringWriter);
+    Mockito.when(response.getWriter()).thenReturn(printWriter);
+  }
+
+  @After
+  public void clear() {
```
antmar, line 82: Prefer to use `tearDown` for the method annotated with `@After`. It's a standard naming convention in JUnit and related frameworks.

Also, you don't need to clear the stringWriter, because it's an instance variable, and JUnit will create a new instance of your class to run each test method.

src/test/java/AuthenticatedHttpServletTest.java
antmar: I agree that you don't need to document every test case. There are some cases where a test is particularly complicated or the condition being tested is not obvious, but this test case in particular is pretty trivial.

(This is a good exception to "public methods need comments".)

src/test/java/AuthenticatedHttpServletTest.java
```diff
@@ -0,0 +1,172 @@
...
+public class AuthenticatedHttpServletTest {
+  // Mock abstract servlet with real constructor and real methods
+  private static final AuthenticationVerifier authVerifier =
+      Mockito.mock(AuthenticationVerifier.class);
+  private static final AuthenticatedHttpServlet servlet =
+      Mockito.mock(
+          AuthenticatedHttpServlet.class,
+          Mockito.withSettings()
+              .useConstructor(authVerifier)
+              .defaultAnswer(Mockito.CALLS_REAL_METHODS));
```
antmar, line 42: It's very weird to see a mock used for the Subject Under Test (SUT). I realize it's an abstract class, so you can't instantiate it directly, but I think it would be more appropriate to make an actual derived class(es) for testing. (You could make them static child classes in this test class, to avoid adding more files.)

One thing you are not really testing is calling the `doGet` and `doPost` in the derived class. I would prefer to have most tests use a derived class that supports both methods (returns a 200 response). Add separate test cases for the default implementations of `doGet` and `doPost` in AuthenticatedHttpServlet.

src/test/java/ClientIDServletTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Before
+  public void init() throws IOException {
```
antmar, line 41: Similar comment about using the canonical name `setUp`

src/test/java/ClientIDServletTest.java
```diff
@@ -0,0 +1,62 @@
...
+    request = Mockito.mock(HttpServletRequest.class);
+    response = Mockito.mock(HttpServletResponse.class);
+
+    stringWriter = new StringWriter();
+    printWriter = new PrintWriter(stringWriter);
+    Mockito.when(response.getWriter()).thenReturn(printWriter);
+  }
+
+  @After
+  public void clear() {
```
antmar, line 51: Similar comment - you don't need to clean up here.

austinteshuba: Removed!

src/test/java/ClientIDServletTest.java
austinteshuba: Fixed!

src/test/java/AuthenticatedHttpServletTest.java
austinteshuba: I understand what you mean, but I'm not sure if there's a functional difference between the two. While this seems like a regular mock, this is a feature of Mockito that actually allows you to use the real methods in the abstract class directly  [Stack Overflow](https://stackoverflow.com/Questions/1087339/using-mockito-to-test-abstract-classes). In all likelihood, Mockito is doing exactly what you're suggesting under the hood. 

In terms of testing the derived tests, they are currently covered though the expected result is a 400 rather than a 200. While making a derived class would allow me to test for a 200 error, it wouldn't add much value to the actual test. I also see value in testing the original doGet and doPost here since that's what Java will call first - it's important to ensure the link between the original method called by Java and the new overloaded method remains intact. 

There's a good chance I'm just not seeing something or misunderstanding your comment (I'm still brand new to this!), but I don't think there's an actual lapse in code coverage here.

src/test/java/AuthenticatedHttpServletTest.java
austinteshuba: Removed this block, but noted!

src/test/java/AuthenticatedHttpServletTest.java
austinteshuba: Cool - changed!

src/main/java/com/google/sps/utility/ServletUtility.java
austinteshuba: Fair enough - I created a custom error type for this

src/main/java/com/google/sps/model/AuthenticationVerifierImpl.java
austinteshuba: Nope, you only need this for verification. 

Theoretically, I could probably get rid of this entire functionality. However, it seemed like a good idea to make sure the user did have some type of valid credential (I have no way of verifying an accessToken's validity, though passing an invalid access token would result in failure). We would 100% need this if we implemented a session store or something along those lines (which is probably why I included it as I initially thought we would go down that route). [Learn more here](https://developers.google.com/identity/sign-in/web/backend-auth). 

I can remove it if you think it would be better, but I don't think additional security (however weak) hurts

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: Interesting! Thanks

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: Fixed same way as above!

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: Because I'm overriding doGet and doPost, they are originally declared with throwing IOException. Java won't let me add to this list of errors thrown in an override, and I cannot change the original definition. I also have to either catch or throw this error (and add it to the method signature) or I'll get a compiler error. 

An IOException (or any unchecked exception) will default to a 500, but I have to catch the GeneralSecurityException. I can. however, throw a ServletException, which is designed for use when the request cannot be handled. 

I just changed the error message, removed the catch expression for the IOException (as this can be thrown from a doGet or doPost), and threw the new error type.

austinteshuba:
Removed all null references from my newly created code (namely with getCookie and getGoogleCredential).

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -93,4 +132,43 @@ public void doPost(
...
+   *     present)
+   */
+  private Credential getGoogleCredential(HttpServletRequest request)
+      throws TokenVerificationException, GeneralSecurityException, IOException,
+          CookieParseException {
+    // If cookies not present, return null (user's session likely expired)
+    Cookie idTokenCookie = ServletUtility.getCookie(request, "idToken");
+    Cookie accessTokenCookie = ServletUtility.getCookie(request, "accessToken");
+    if (idTokenCookie == null || accessTokenCookie == null) {
+      return null;
```
tttzach, line 156: Would this merit an exception to be thrown?

src/main/java/com/google/sps/utility/ServletUtility.java
```diff
@@ -65,8 +68,10 @@ public static Cookie getCookie(HttpServletRequest request, String cookieName) {
    * @param request contains cookies for a userToken and accessToken
    * @return Value of the "Authorization" header. Null if authentication is invalid or accessToken
    *     was not found
+   * @throws CookieParseException if duplicate cookies are found and, thus, a value cannot be parsed
    */
-  public static String generateAuthorizationHeader(HttpServletRequest request) {
+  public static String generateAuthorizationHeader(HttpServletRequest request)
+      throws CookieParseException {
     // If accessToken cannot be found, return null
```
tttzach, line 75: Would this merit an exception to be thrown?

src/main/java/com/google/sps/model/AuthenticationVerifier.java
```diff
@@ -0,0 +1,35 @@
...
+package com.google.sps.model;
+
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+/** Contract for verifying user with Google Sign in API */
+public interface AuthenticationVerifier {
+  // OAuth 2.0 Client ID
+  // TODO: Hide Client ID (Issue #72)
+  String CLIENT_ID = "12440562259-mf97tunvqs179cu1bu7s6pg749gdpked.apps.googleusercontent.com";
```
tttzach, line 24: I'll handle this!

src/main/java/com/google/sps/utility/ServletUtility.java
```diff
@@ -27,53 +28,69 @@
...
+   * @param cookieName name of desired cookie. Case sensitive
+   * @return true if present (duplicates will also return true), false otherwise
+   */
+  public static boolean hasCookie(HttpServletRequest request, String cookieName) {
+    List<Cookie> cookies =
+        Arrays.stream(request.getCookies())
+            .filter((Cookie c) -> c.getName().equals(cookieName))
+            .collect(Collectors.toList());
+
+    return cookies.size() > 0;
```
tttzach, line 76: extreme nit: Here's a [debate](https://stackoverflow.com/questions/9341740/ifliststr-size-0-versus-ifliststr-isempty/9341835) between using `.size() > 0` and `.isEmpty()`. Don't think it really matters though.

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
tttzach: This comment is no longer relevant since I reviewed outdated code.

src/main/java/com/google/sps/utility/ServletUtility.java
tttzach: This comment is no longer relevant since I reviewed outdated code.

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -129,29 +136,34 @@ public void doPost(
...
-    if (idTokenCookie == null || accessTokenCookie == null) {
-      return null;
+      throws CredentialVerificationException, GeneralSecurityException, IOException {
+    Cookie idTokenCookie;
+    Cookie accessTokenCookie;
+
+    try {
+      idTokenCookie = ServletUtility.getCookie(request, "idToken");
+    } catch (CookieParseException e) {
+      throw new CredentialVerificationException("idToken is not present / cannot be parsed!");
```
antmar, line 152: Whenever you catch an exception and throw a different exception, you should make the original exception the "cause" or "inner exception". All built-in exception classes should have a constructor that include an Exception argument after the message string. Your own exceptions should have these constructors as well.

austinteshuba: This is cool! Implemented this no problem - seems helpful for keeping a chain of how exceptions were caused (which I always benefit from when tests fail / I'm debugging). Thanks!

src/main/java/com/google/sps/utility/ServletUtility.java
austinteshuba: Fixed just to keep consistent. I don't know why I keep using .size() instead of .isEmpty() - force of habit I guess!

PR #70, opened 2020-07-13
Refactor gmail to use client abstraction pattern

austinteshuba:
Includes unit tests for GmailServlet.

Dependent on PR #69


src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Get a message from a user's Gmail account
+   *
+   * @param messageId the messageID (retrieved from listUserMessages) of the desired Message
+   * @param format MessageFormat enum that defines how much of the Message object is populated
+   * @return a Message object with the requested information
+   * @throws IOException if an issue occurs with the Gmail service
+   */
+  Message getUserMessage(String messageId, MessageFormat format) throws IOException;
```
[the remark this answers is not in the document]
austinteshuba: Technically, I think I'm actually supposed to put them on the interface. 

Overrides are an exception to the rules in the style guide, but interfaces are not. I'm going to go back and fix instances of these issues

src/test/java/GmailServletTest.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+@RunWith(JUnit4.class)
+public final class GmailServletTest {
+  private static final AuthenticationVerifier authenticationVerifier =
+      Mockito.mock(AuthenticationVerifier.class);
+  private static final GmailClientFactory gmailClientFactory =
+      Mockito.mock(GmailClientFactory.class);
+  private static final GmailClient gmailClient = Mockito.mock(GmailClient.class);
+  private static final GmailServlet servlet =
+      new GmailServlet(authenticationVerifier, gmailClientFactory);
```
gbuenoandrade, line 51: I think it's safer to initialize all these fields inside `init()` and make them non-static for the reasons I mentioned in the previous PR (or inside `classInit` when appropriate to make it clear that they will not be changing throughout tests)

src/test/java/GmailServletTest.java
```diff
@@ -0,0 +1,119 @@
...
+  private HttpServletRequest request;
+  private HttpServletResponse response;
+  private StringWriter stringWriter;
+  private PrintWriter printWriter;
+
+  private static final String MESSAGE_ID_ONE = "messageIdOne";
+  private static final String MESSAGE_ID_TWO = "messageIdTwo";
+  private static final String MESSAGE_ID_THREE = "messageIdThree";
+  private static final List<Message> noMessages = new ArrayList<>();
+  private static final List<Message> threeMessages =
```
gbuenoandrade, line 73: Nit: [ImmutableList.of()](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/ImmutableList.html#asList--)

src/test/java/GmailServletTest.java
```diff
@@ -0,0 +1,119 @@
...
+      Arrays.asList(
+          new Message().setId(MESSAGE_ID_ONE),
```
gbuenoandrade, line 75: FYI: If `Message` is immutable and you use `ImmutableList.of`, this will effectively become a constant

src/test/java/GmailServletTest.java
```diff
@@ -0,0 +1,119 @@
...
+  private HttpServletResponse response;
+  private StringWriter stringWriter;
+  private PrintWriter printWriter;
+
+  private static final String MESSAGE_ID_ONE = "messageIdOne";
+  private static final String MESSAGE_ID_TWO = "messageIdTwo";
+  private static final String MESSAGE_ID_THREE = "messageIdThree";
+  private static final List<Message> noMessages = new ArrayList<>();
+  private static final List<Message> threeMessages =
+      Arrays.asList(
```
gbuenoandrade, line 74: Nit: ImmutableList.of()

austinteshuba: Fixed! Thanks

src/test/java/GmailServletTest.java
austinteshuba: Awesome! Thanks

src/test/java/GmailServletTest.java
austinteshuba: Oh cool - I forgot about this. Thanks!

src/test/java/GmailServletTest.java
austinteshuba: Fixed!

src/main/java/com/google/sps/utility/GmailUtility.java
```diff
@@ -14,44 +14,10 @@
...
-import com.google.api.services.gmail.model.Message;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
-
-/**
- * Contains logic that handles GET & POST requests to the Gmail API and transforms those responses
- * into easily usable Java types
- */
+/** Contains logic that transforms and creates Gmail data structures (no network requests) */
```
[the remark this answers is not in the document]
austinteshuba: They will always relate to the GmailClient (since they are used for listing emails). As such, these are probably better as public static classes declared in the GmailClient interface (they semantically belong to the GmailClient, though they could be static since they don't operate on any specific instance - they belong to the class).

src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Encapsulates possible values for the "format" query parameter in the Gmail GET message method
+   * FULL: Returns full email message data METADATA: Returns only email message ID, labels, and
+   * email headers MINIMAL: Returns only email message ID and labels; does not return the email
+   * headers, body, or payload. RAW: Returns the full email message data with body content in the
+   * raw field as a base64url encoded string;
```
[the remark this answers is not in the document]
austinteshuba: I had no clue this was possible! I like this a lot - thanks

src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.api.services.gmail.model.Message;
+import java.io.IOException;
+import java.util.List;
+
+/** Contract for trivial get requests from the Gmail API */
+public interface GmailClient {
+  /**
+   * Get all of the message IDs from a user's Gmail account
+   *
+   * @param query search query to filter which results are returned
```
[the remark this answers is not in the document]
austinteshuba: Good call - fixing!

src/test/java/GmailServletTest.java
```diff
@@ -0,0 +1,114 @@
...
+import org.mockito.Mockito;
+
+/**
+ * Test Gmail Servlet to ensure response contains correctly parsed messageIds. Assumes
+ * AuthenticatedHttpServlet is functioning properly (those tests will fail otherwise).
+ */
+@RunWith(JUnit4.class)
+public final class GmailServletTest {
+  private GmailClient gmailClient;
+  private GmailServlet servlet;
```
antmar, line 44: Group instance and static variables together

src/test/java/GmailServletTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    // Writer used in get/post requests to capture HTTP response values
+    stringWriter = new StringWriter();
+    printWriter = new PrintWriter(stringWriter);
+    Mockito.when(response.getWriter()).thenReturn(printWriter);
+  }
+
+  @Test
+  public void noMessagesPresent() throws IOException, ServletException {
+    Mockito.when(gmailClient.listUserMessages(Mockito.anyString())).thenReturn(NO_MESSAGES);
```
antmar, line 101: Can you restrict the mock to return the result only when the correct query is provided? That way, you'll be testing more parts of your code. Same thing below.

src/test/java/GmailServletTest.java
austinteshuba: Agreed - fixed!

src/test/java/GmailServletTest.java
austinteshuba: Agreed - fixed! Thanks for catching this

PR #80, opened 2020-07-16
Refactor calendar utility

AymarMuhikira:
New Implementation that allows for unit testing. Replaces PR #57


src/test/java/CalendarServletTest.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  @Test
+  public void allEvent() throws IOException, ServletException {
+    // Test case where there are two defined and an undefined event
+    Mockito.when(calendarClient.getCalendarEvents()).thenReturn(EVENT_ALL);
+    servlet.doGet(request, response);
+    printWriter.flush();
+    Assert.assertTrue(stringWriter.toString().contains(EVENT_ALL_JSON));
+  }
```
tttzach, line 139: nit: It might help to decode the JSON: https://github.com/googleinterns/step7-2020/pull/40#discussion_r453738090

src/test/java/CalendarServletTest.java
```diff
@@ -0,0 +1,140 @@
...
+  private static final Cookie sampleIdTokenCookie = new Cookie(ID_TOKEN_KEY, ID_TOKEN_VALUE);
+  private static final Cookie sampleAccessTokenCookie =
+      new Cookie(ACCESS_TOKEN_KEY, ACCESS_TOKEN_VALUE);
+  private static final Cookie[] validCookies =
+      new Cookie[] {sampleIdTokenCookie, sampleAccessTokenCookie};
+
+  private HttpServletRequest request;
+  private HttpServletResponse response;
+  private StringWriter stringWriter;
+  private PrintWriter printWriter;
```
tttzach, line 59: extreme nit: maybe move this up with the AuthenticationVerifier, CalendarClientFactory, etc. just for readability

src/test/java/CalendarServletTest.java
AymarMuhikira: Agreed, I changed the tests to decode instead

src/test/java/CalendarServletTest.java
AymarMuhikira: Agreed. Fixed

src/test/java/CalendarServletTest.java
```diff
@@ -0,0 +1,155 @@
...
+  private CalendarClientFactory calendarClientFactory;
+  private CalendarClient calendarClient;
+  private CalendarServlet servlet;
+  private HttpServletRequest request;
+  private HttpServletResponse response;
+  private StringWriter stringWriter;
+  private PrintWriter printWriter;
+  private final Gson gson = new Gson();
+
+  private static final boolean AUTHENTICATION_VERIFIED = true;
```
gbuenoandrade, line 53: Nit: I think it is simpler to just use `true` directly

src/test/java/CalendarServletTest.java
```diff
@@ -0,0 +1,155 @@
...
+  private static final String ID_TOKEN_VALUE = "sampleId";
+  private static final String ACCESS_TOKEN_KEY = "accessToken";
+  private static final String ACCESS_TOKEN_VALUE = "sampleAccessToken";
+  private static final Cookie sampleIdTokenCookie = new Cookie(ID_TOKEN_KEY, ID_TOKEN_VALUE);
+  private static final Cookie sampleAccessTokenCookie =
+      new Cookie(ACCESS_TOKEN_KEY, ACCESS_TOKEN_VALUE);
+  private static final Cookie[] validCookies =
+      new Cookie[] {sampleIdTokenCookie, sampleAccessTokenCookie};
+
+  // Events must be returned in order of retrieval - JSON includes tasks in desired order
```
gbuenoandrade, line 64: Nit: please move this comment to the tests where it applies

src/test/java/CalendarServletTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+  @Test
+  public void noCalendarEvent() throws IOException, ServletException {
+    // Test case where there are no events in the user's calendar
+    Mockito.when(calendarClient.getCalendarEvents()).thenReturn(NO_EVENT);
+    servlet.doGet(request, response);
+    printWriter.flush();
+    String actualString = stringWriter.toString();
+    Type type = new TypeToken<List<Event>>() {}.getType();
+    List<Event> actual = gson.fromJson(actualString, type);
```
gbuenoandrade, line 116: Maybe turn this into an auxiliary method to avoid code repetition?

src/main/java/com/google/sps/model/CalendarClient.java
```diff
@@ -0,0 +1,30 @@
...
+
+/** Interface to handle get requests to the Calendar API. */
+public interface CalendarClient {
+  /**
+   * Get the events in a user's calendar.
+   *
+   * @return the list of Event from the calendar
+   * @throws IOException thrown when an issue occurs
+   */
+  List<Event> getCalendarEvents() throws IOException;
```
antmar, line 29: Fly-by comment - it would be good to pass in the calendar/list to retrieve events from, even if you always ask for "primary" when you call this method. We know that eventually we want to specify the calendar(s), so let's make the API support it.

It would also be more consistent with TaskClient and GmailClient.

src/test/java/CalendarServletTest.java
```diff
@@ -40,17 +40,16 @@
...
-  private final Gson gson = new Gson();
+  private static AuthenticationVerifier authenticationVerifier;
+  private static CalendarClientFactory calendarClientFactory;
+  private static CalendarClient calendarClient;
+  private static CalendarServlet servlet;
+  private static HttpServletRequest request;
+  private static HttpServletResponse response;
+  private static StringWriter stringWriter;
+  private static PrintWriter printWriter;
+  private static final Gson gson = new Gson();
```
gbuenoandrade, line 51: Nit: I'm sorry for the confusion. I meant only the gson field should be static. The others you are better off leaving non-static to avoid scenarios like the following possible sequence of calls performed by JUnit:
```
Test test1 = Test();
Test test2 = Test();
test1.init()
test2.init()
test1.test1()
# your static fields are not necessarily in the same state they were after the init() call
test2.test2()
```

(The gson is fine because it won't change throughout the tests.)

src/test/java/CalendarServletTest.java
```diff
@@ -109,47 +106,39 @@ public void setUp() throws IOException, GeneralSecurityException {
...
+
+  public void checkReturn(List<Event> EXPECTED_EVENT) throws IOException, ServletException {
+    // Method that handles the once the Calendar Client has been mocked
     servlet.doGet(request, response);
     printWriter.flush();
     String actualString = stringWriter.toString();
     Type type = new TypeToken<List<Event>>() {}.getType();
     List<Event> actual = gson.fromJson(actualString, type);
-    Assert.assertEquals(EVENT_ALL, actual);
+    Assert.assertEquals(EXPECTED_EVENT, actual);
```
gbuenoandrade, line 142: Super nit: I'd rather leave the assertion outside the helper method so it is easier to debug. Knowing that the assertion inside the helper method failed isn't enough to know which exact test failed (you would still need to check the stack trace). Having the assertions inside the tests remove that need.

src/test/java/CalendarServletTest.java
```diff
@@ -105,40 +116,49 @@ public void setUp() throws IOException, GeneralSecurityException {
...
-    Mockito.when(calendarClient.getCalendarEvents()).thenReturn(EVENT_ALL);
-    checkReturn(EVENT_ALL);
+    Mockito.when(calendarClient.getCalendarList()).thenReturn(ONE_CALENDAR);
+    Mockito.when(calendarClient.getCalendarEvents(PRIMARY)).thenReturn(EVENT_ALL);
+    List<Event> actual = checkReturn();
+    Assert.assertEquals(EVENT_ALL, actual);
   }
 
-  public void checkReturn(List<Event> EXPECTED_EVENT) throws IOException, ServletException {
+  public List<Event> checkReturn() throws IOException, ServletException {
```
antmar, line 155: Nit: rename this helper to something like `doGetAsJSON` or equally descriptive. (It's not checking anything now.)

(Also, it doesn't need to be public.)

Issue #43, opened 2020-07-08
MVP Infrastructure: Gmail panel


tttzach:
Populate Gmail panel with relevant data as depicted under 'Requirements & Scale' in our design document.

Issue #85, opened 2020-07-17
UI: vertical-center should be used in the enclosing div, not on text elements themselves


austinteshuba:
By using `u-vertical-center` on text elements directly, spans act strangely and require `&nbsp` to look close to specification. Instead, use this utility on the enclosing div - it will have the same effect and this behavior will be avoided.

Issue #84, opened 2020-07-16
Gmail Panel: Use batching or multithreading to compute the gmail statistics faster


austinteshuba:
Right now, a single thread is used to get the number of unread emails, number of important unread emails, number of recent unread emails, and most frequent sender of unread emails. This is slow. 

Look into using multithreading and/or batching of requests. 

Past progress / attempts / learnings: 
Note that the GmailService.Builder is not a thread-safe implementation, and new threads have to be made with the ThreadManager API to be used with appengine. Be sure to check the responses from the servlets to ensure they have actual statistics (not just 0's and "", as these represent null values and may not reflect what should actually be returned). Making a new GmailService instance in each branch did not work during first attempt.

Issue #83, opened 2020-07-16
Gmail Panel: Accept query parameters for days and hours


austinteshuba:
User should be able to set how many days of emails we should be checking the unread/important/most frequent sender statistics for. Similarly, User should be able to select how many hours of emails we are checking the recent unread emails statistics for.

austinteshuba:
The server has support and tests for this. All that needs to be implemented is that the client has to have the option to send the parameters to server.

PR #86, opened 2020-07-17
Create MVP version of Gmail Panel

austinteshuba:
Gmail panel now includes all desired statistics 
Screenshot 2020-07-16 at 8 39 19 PM

There are some identified issues that need to be completed at a later data. These include making UI elements so the client can send query parameters to server (Issue #83), using multithreading to improve performance (Issue #84), as well as some small style discrepancies (Issue #85)

Closes issue #43

Dependent on PR #93


src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -76,20 +76,22 @@
    *     ignore filter
    * @param emailAgeUnits "d" for days, "h" for hours, "" for ignore email
    * @param unreadOnly true if only returning unread emails, false otherwise
+   * @param isImportant true if only returning important emails, false otherwise
    * @param from email address of the sender. "" if not specified
    * @return string to use in gmail (either client or API) to find emails that match criteria
    */
   static String emailQueryString(
-      int emailAge, String emailAgeUnits, boolean unreadOnly, String from) {
+      int emailAge, String emailAgeUnits, boolean unreadOnly, boolean isImportant, String from) {
```
tttzach, line 84: nit: I should've mentioned this in the PR this was introduced. Use `StringBuilder` rather than [creating a new string](https://stackoverflow.com/questions/5234147/why-stringbuilder-when-there-is-string) for each parameter.

src/main/java/com/google/sps/model/GmailResponse.java
```diff
@@ -0,0 +1,79 @@
...
+      int unreadEmailsHours,
+      int unreadImportantEmails,
+      String sender) {
+    this.nDays = nDays;
+    this.mHours = mHours;
+    this.unreadEmailsDays = unreadEmailsDays;
+    this.unreadEmailsHours = unreadEmailsHours;
+    this.unreadImportantEmails = unreadImportantEmails;
+    this.sender = sender;
+  }
```
tttzach, line 54: Conceptual question - does Java have an equivalent to C++'s Member Initialization List where you could make the constructor a one liner `GmailResponse(int nDays) : nDays{nDays} {}`

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -66,15 +74,131 @@ public void doGet(
...
+  /**
+   * Get the amount of unread emails from the last n days in a user's Gmail account
+   *
+   * @param nDays number of days to check the user's account for
+   * @param gmailClient Gmail service with valid google credential
+   * @return number of unread emails from last nDays days
+   * @throws IOException if an issue occurs with the Gmail Service
+   */
+  private int getUnreadEmailCountDays(int nDays, GmailClient gmailClient) throws IOException {
+    String searchQuery = GmailClient.emailQueryString(nDays, "d", true, false, "");
```
tttzach, line 118: Might be worth explaining what `"d"`, `true`, `false` and `""` represent in `.emailQueryString` or assigning them to constants like you did in GmailClientTest.

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -66,15 +74,131 @@ public void doGet(
...
+          String sender = senderList.get(0).getValue();
+
+          senders.put(sender, senders.get(sender) != null ? senders.get(sender) + 1 : 1);
+        });
+    String sender = senders.entrySet().stream().max(Map.Entry.comparingByValue()).get().getKey();
+
+    if (sender.charAt(0) == '<') {
+      return sender.substring(1, sender.length() - 1);
+    } else {
+      return sender.split("<")[0].trim();
```
tttzach, line 201: Perhaps a comment to show how `sender` is supposed to typically look, hard to imagine what these operations would do to `sender`.

src/main/webapp/index.html
```diff
@@ -46,32 +47,30 @@ <h2 class="panel-row__header u-align-with-first-panel u-text-shadow">Services (a
...
               <div class="panel__rounded-number u-background-mail">
-                <p class="panel__text-data-point u-text-white">35
+                <p class="panel__text-data-point u-text-white" id="gmailUnreadEmailsDays">
               </div>
               <p class="panel__text-data-point u-vertical-center">Unread Emails:
             </div>
             <div class="panel__content-entry">
               <img class="panel__content-icon" alt="clock-icon" src="assets/clock.png">
-              <p class="panel__text-large u-vertical-center">2 from last 3 hours
+              <p class="panel__text-large u-vertical-center"><span id="gmailUnreadEmailsHours"></span>&nbsp;from the last&nbsp;<span id="gmailMHours"></span>&nbsp;hours
```
tttzach, line 60: I could be wrong but it seems like `&nbsp` is used more for styling purposes. Is there a way to abstract this into CSS?

austinteshuba:
To next reviewers, I would like your two cents on this: 

Theoretically, I should be able to get the number of unread emails from last nDays, number of unread emails from last mHours, and number of unread, important emails from the last nDays, at the same time as finding the most frequent sender of unread emails. I could do this while iterating through the emails sent from gmail (which I need to do to find the most frequent sender of unread emails).

This would save on API calls, but this would make things harder to test / make multi-threaded in the future. I kept them separated for now, but I'm wondering if the slight gain in efficiency is worth creating tightly coupled code. As of now I'm tending towards keeping the methods separate, but does anyone have strong feelings either way / ideas about how to improve efficiency without sacrificing code quality?

src/main/webapp/index.html
austinteshuba: `&nbsp` has a unique meaning in HTML, and cannot be abstracted. They should, however, be avoided, but this was a required workaround based on a style mistake I made in my UI pull request. This is corrected in the next dependent PR, #87 .

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: There are comments regarding this on lines 176 to 182. I changed the variable name to `headerValue` so the connection between this variable and that comment is a bit more intuitive.

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: This is documented in the `GmailClient.emailQueryString()` javadoc. I think it might be redundant to repeat this here.

src/main/java/com/google/sps/model/GmailResponse.java
austinteshuba: Looks like the short answer is [no](https://stackoverflow.com/questions/14664374/is-there-member-initialization-list-syntax-in-java-like-c).

src/main/java/com/google/sps/model/GmailClient.java
austinteshuba: Fixed this in a separate PR (#92). Thanks!

gbuenoandrade:
> Dependent on PR #93

Why is this dependent on that PR? The changes are unrelated.

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -66,15 +74,132 @@ public void doGet(
...
+    HashMap<String, Integer> senders = new HashMap<>();
+    unreadEmails.forEach(
+        (Message m) -> {
+          List<MessagePartHeader> senderList = m.getPayload().getHeaders();
+          senderList =
+              senderList.stream()
+                  .filter((MessagePartHeader header) -> header.getName().equals("From"))
+                  .collect(Collectors.toList());
+
+          String sender = senderList.get(0).getValue();
```
gbuenoandrade, line 192: [findFirst](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#findFirst--)

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -66,15 +74,132 @@ public void doGet(
...
+          senders.put(sender, senders.get(sender) != null ? senders.get(sender) + 1 : 1);
+        });
+    String headerValue =
+        senders.entrySet().stream().max(Map.Entry.comparingByValue()).get().getKey();
+
+    if (headerValue.charAt(0) == '<') {
+      return headerValue.substring(1, headerValue.length() - 1);
+    } else {
+      return headerValue.split("<")[0].trim();
+    }
```
gbuenoandrade, line 203: Nit: maybe a ternary if?

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -66,15 +74,132 @@ public void doGet(
...
+                  try {
+                    return gmailClient.getUserMessage(m.getId(), messageFormat);
+                  } catch (IOException e) {
+                    throw new RuntimeException(e);
+                  }
+                })
+            .collect(Collectors.toList());
+    if (unreadEmails.isEmpty()) {
+      return "";
+    }
```
gbuenoandrade, line 174: I wonder if it would make more sense for the auxiliary methods to live in `GmailClient`

src/test/java/GmailClientTest.java
```diff
@@ -119,14 +124,31 @@ public void getQueryStringFromAnySender() {
     Assert.assertEquals(fromAnySenderQuery, EMPTY_QUERY);
   }
 
+  @Test
+  public void getQueryStringImportantEmailsOnly() {
+    String importantQuery = GmailClient.isImportantQuery(RETURN_IMPORTANT_ONLY);
```
gbuenoandrade, line 129: Nit: I think `true` would be clearer

src/test/java/GmailClientTest.java
```diff
@@ -119,14 +124,31 @@ public void getQueryStringFromAnySender() {
...
+
+    Assert.assertEquals(importantQuery, IMPORTANT_EMAILS_QUERY);
+  }
+
+  @Test
+  public void getQueryStringDontFilterForImportantTag() {
+    String noImportantFilterQuery = GmailClient.isImportantQuery(DEFAULT_IMPORTANT_FILTER);
```
gbuenoandrade, line 136: Nit: I think `false` would be clearer

src/test/java/GmailServletTest.java
```diff
@@ -71,12 +69,79 @@
...
+                                  .setValue(SENDER_WITHOUT_NAME)))));
+  private static final List<Message> THREE_MESSAGES_MAJORITY_SENDER_WITHOUT_NAME =
+      ImmutableList.of(
+          new Message()
+              .setId(MESSAGE_ID_ONE)
+              .setPayload(
+                  new MessagePart()
+                      .setHeaders(
+                          Collections.singletonList(
+                              new MessagePartHeader().setName("From").setValue(SENDER_WITH_NAME)))),
```
gbuenoandrade, line 126: Nit: Auxiliary method that takes an id and a name to avoid code repetition?

src/test/java/GmailServletTest.java
```diff
@@ -101,22 +166,277 @@ public void setUp() throws IOException, GeneralSecurityException {
...
+    Mockito.when(gmailClient.getUserMessage(Mockito.contains(MESSAGE_ID_ONE), Mockito.any()))
+        .thenReturn(THREE_MESSAGES_MAJORITY_SENDER_WITH_NAME.get(0));
+    Mockito.when(gmailClient.getUserMessage(Mockito.contains(MESSAGE_ID_TWO), Mockito.any()))
+        .thenReturn(THREE_MESSAGES_MAJORITY_SENDER_WITH_NAME.get(1));
+    Mockito.when(gmailClient.getUserMessage(Mockito.contains(MESSAGE_ID_THREE), Mockito.any()))
+        .thenReturn(THREE_MESSAGES_MAJORITY_SENDER_WITH_NAME.get(2));
+
+    servlet.doGet(request, response);
+    printWriter.flush();
+    GmailResponse gmailResponse = gson.fromJson(stringWriter.toString(), GmailResponse.class);
```
gbuenoandrade, line 337: I think an auxiliary method for these calls would help

src/test/java/GmailServletTest.java
```diff
@@ -101,22 +166,277 @@ public void setUp() throws IOException, GeneralSecurityException {
...
+    Mockito.when(request.getParameter(Mockito.eq("nDays")))
+        .thenReturn(String.valueOf(NEGATIVE_N_DAYS));
+    Mockito.when(request.getParameter(Mockito.eq("mHours")))
+        .thenReturn(String.valueOf(DEFAULT_M_HOURS));
+
+    Mockito.when(gmailClient.listUserMessages(Mockito.anyString())).thenReturn(NO_MESSAGES);
+
+    servlet.doGet(request, response);
+    printWriter.flush();
+    Mockito.verify(response, Mockito.times(1)).sendError(Mockito.eq(400), Mockito.anyString());
```
gbuenoandrade, line 205: Nit: It's preferred to test state than interactions. If `Mockito` makes that hard, a fake could be an option.

gbuenoandrade:
> To next reviewers, I would like your two cents on this:
> 
> Theoretically, I should be able to get the number of unread emails from last nDays, number of unread emails from last mHours, and number of unread, important emails from the last nDays, at the same time as finding the most frequent sender of unread emails. I could do this while iterating through the emails sent from gmail (which I need to do to find the most frequent sender of unread emails).
> 
> This would save on API calls, but this would make things harder to test / make multi-threaded in the future. I kept them separated for now, but I'm wondering if the slight gain in efficiency is worth creating tightly coupled code. As of now I'm tending towards keeping the methods separate, but does anyone have strong feelings either way / ideas about how to improve efficiency without sacrificing code quality?

If you could come up with a query such that the result `gmailClient.listUserMessages(query)` could be reused across the calculation of the values you need, you could then just pass that to `getUnreadEmailCountDays`, `getUnreadEmailCountHours`, `getImportantUnreadEmailCount`  and  `getMostFrequentSender` to save a few API calls. I think that would be a win. I don't see any real damage to code quality being cause by that.

austinteshuba:
@austinteshuba note to self - handle the case where there are two+ equally frequent senders (e.g. google sends 2, Sam sends 2, who gets returned?) 

Some ideas can include returning both to the client, or maybe just the one who sent the most recent emails.

src/main/java/com/google/sps/model/GmailResponse.java
```diff
@@ -0,0 +1,79 @@
...
+ * methods to generate these statistics. Properties only available through reflective access (i.e.
+ * gson, etc)
+ */
+public class GmailResponse {
+  private int nDays;
+  private int mHours;
+  private int unreadEmailsDays;
+  private int unreadEmailsHours;
+  private int unreadImportantEmails;
+  private String sender;
```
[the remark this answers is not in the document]
austinteshuba: Agreed - fixed!

src/main/java/com/google/sps/model/GmailResponse.java
```diff
@@ -0,0 +1,79 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.model;
+
+/**
+ * Contains the summary gmail information that should be passed to the client, as well as the
+ * methods to generate these statistics. Properties only available through reflective access (i.e.
+ * gson, etc)
```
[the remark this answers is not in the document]
austinteshuba: Agreed - fixed!

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -66,15 +74,132 @@ public void doGet(
...
+    List<Message> unreadEmails;
+    unreadEmails =
+        gmailClient.listUserMessages(searchQuery).stream()
+            .map(
+                (Message m) -> {
+                  try {
+                    return gmailClient.getUserMessage(m.getId(), messageFormat);
+                  } catch (IOException e) {
+                    throw new RuntimeException(e);
+                  }
```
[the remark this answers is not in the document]
austinteshuba: You cannot throw checked exceptions in a stream, SO i just wrapped it in a RuntimeException for simplicity. [Handling checked exceptions in Java streams](https://www.oreilly.com/content/handling-checked-exceptions-in-java-streams/)

src/main/java/com/google/sps/servlets/GmailServlet.java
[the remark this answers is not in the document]
austinteshuba: I actually want to the value in the response to be "", so the client knows that there is no most frequent sender (because there are no unread emails). 

I'm hesitant to throw an exception because this isn't really an exceptional path - having no unread emails from the last N days is not erroneous (though surprising)

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: Cool - thanks for this!

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: Done!

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: I actually think they would do better in the GmailResponse class, since that class has to do with the actual Gmail statistics. GmailClient is more for POST/GET requests to and from the Gmail service and any methods that govern those transactions. 

I moved them over to GmailResponse - let me know what you think!

src/test/java/GmailClientTest.java
austinteshuba: Just for consistency, I think it might be better to continue using constants for every value (we agreed to this in a standup a while back). 

I don't mind changing this though if you have strong feelings either way. Please let me know!

src/test/java/GmailServletTest.java
austinteshuba: Done!

src/test/java/GmailServletTest.java
austinteshuba: Agreed - fixed!

src/test/java/GmailServletTest.java
austinteshuba: I gave the fake a try - thanks for the idea! Let me know if it looks good/bad.

src/test/java/GmailClientTest.java
austinteshuba: Just for consistency, I think it might be better to continue using constants for every value (we agreed to this in a standup a while back).

I don't mind changing this though if you have strong feelings either way. Please let me know!

src/main/java/com/google/sps/servlets/GmailServlet.java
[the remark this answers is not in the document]
gbuenoandrade: ACK. That's a really odd design decision made by Oracle

src/main/java/com/google/sps/model/GmailResponse.java
```diff
@@ -0,0 +1,288 @@
...
+                    return a;
+                  } else if (b.getValue().getTimesSent() > a.getValue().getTimesSent()) {
+                    return b;
+                  } else {
+                    if (a.getValue().getMostRecentEmailTimestamp()
+                        > b.getValue().getMostRecentEmailTimestamp()) {
+                      return a;
+                    }
+                    return b;
+                  }
```
gbuenoandrade, line 274: Try to reduce branching. Maybe something like the following could be a bit cleaner:
```
if (a.getValue().getTimesSent() != b.getValue().getTimesSent()) {
  return a.getValue().getTimesSent() > b.getValue().getTimesSent() ? a : b;
}
return a.getValue().getMostRecentEmailTimestamp() > b.getValue().getMostRecentEmailTimestamp() ? a : b;
```

src/main/java/com/google/sps/model/GmailResponse.java
```diff
@@ -0,0 +1,288 @@
...
+
+  /**
+   * Get the sender of the most unread emails from the last nDays days in a user's Gmail account
+   *
+   * @param unreadEmails a list of unread emails (with METADATA OR FULL MessageFormat)
+   * @return sender of the most unread emails from the last nDays days
+   * @throws GmailMessageFormatException if the messages do not contain an internal date and/or
+   *     headers (and are thus the wrong format)
+   */
+  private String findMostFrequentSender(List<Message> unreadEmails)
```
gbuenoandrade, line 171: API calls are potentially expensive, but just looping through the data is very cheap even for large sets, so don't be afraid of looping through it a couple times if that makes things simpler. I think you could have one method that would just get the most frequent sender and another one that, given the most frequent sender, would find their most recent email.

src/main/java/com/google/sps/model/GmailResponse.java
```diff
@@ -0,0 +1,288 @@
...
+    // last nDays and 2) what the timestamp was of their most recent email.
+    HashMap<String, EmailFrequencyWithDate> senders = new HashMap<>();
+    unreadEmails.forEach(
+        (Message m) -> {
+          long timestamp = m.getInternalDate();
+          MessagePart messagePayload = m.getPayload();
+
+          if (messagePayload == null || timestamp == 0) {
+            throw new GmailMessageFormatException("Messages must be of format METADATA or FULL");
+          }
```
gbuenoandrade, line 233: Can we perform a sanity check on `unreadEmails` only once? Maybe in the constructor to avoid having multiple checks throughout your auxiliary methods.

src/main/java/com/google/sps/model/GmailResponse.java
```diff
@@ -0,0 +1,288 @@
...
+      public void updateMostRecentEmailTimestamp(long timestamp) {
+        if (timestamp > mostRecentEmailTimestamp) {
+          this.mostRecentEmailTimestamp = timestamp;
+        }
+      }
+
+      public void incrementTimesSent() {
+        timesSent++;
+      }
+    }
```
gbuenoandrade, line 213: Although I agree that this class only makes sense inside the method, I think class definitions inside methods hurts readability considerably. I'd rather define this inside `GmailResponse`.

src/test/java/GmailClientTest.java
gbuenoandrade: Raw booleans are already in a way constants mapping true/false to 1/0. I think it'd be fair to use them instead of custom constants even if we are using constants "for every value". But if you do want to define custom ones, maybe rename them so `isImportantQuery(RETURN_IMPORTANT_ONLY)` becomes more "readable"?

src/test/java/HttpServletResponseFake.java
```diff
@@ -0,0 +1,224 @@
...
+   * Get the stringwriter used for writing the response
+   *
+   * @return Stringwriter object
+   */
+  public StringWriter getStringWriter() {
+    printWriter.flush();
+    return stringWriter;
+  }
+
+  @Override
```
gbuenoandrade, line 42: Please leave all the unsupported methods together.

src/test/java/GmailServletTest.java
```diff
@@ -91,32 +182,394 @@ public void setUp() throws IOException, GeneralSecurityException {
...
+   */
+  private GmailResponse getGmailResponse(
+      HttpServletRequest request, HttpServletResponseFake response)
+      throws IOException, ServletException {
+    servlet.doGet(request, response);
+    return gson.fromJson(response.getStringWriter().toString(), GmailResponse.class);
+  }
+
+  @Test
+  public void noNDaysParameter() throws IOException, ServletException {
```
gbuenoandrade, line 237: I was reading our internal doc on Java practices and it says it is okay to simply throw `Exception` from tests

src/main/java/com/google/sps/servlets/GmailServlet.java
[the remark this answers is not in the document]
austinteshuba: Agreed :/

src/main/java/com/google/sps/servlets/GmailServlet.java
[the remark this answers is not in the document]
austinteshuba: Cool idea, I haven't declared one of these before. Thanks!

austinteshuba:
@austinteshuba Move the GmailResponse business logic to a utility (based on feedback from @gbuenoandrade and @antmar )

src/main/java/com/google/sps/exceptions/GmailMessageFormatException.java
```diff
@@ -0,0 +1,40 @@
...
+    super(errorMessage);
+  }
+
+  /**
+   * Creates GmailMessageFormatException instance
+   *
+   * @param errorMessage reason for throwing error
+   * @param exceptionCause initial error that caused the exception
+   */
+  public GmailMessageFormatException(String errorMessage, Exception exceptionCause) {
```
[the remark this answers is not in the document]
austinteshuba: Nice catch - thank you!

src/main/java/com/google/sps/model/GmailResponse.java
```diff
@@ -0,0 +1,288 @@
...
+   * @param unreadEmails a list of Message objects from user's gmail account. Messages must be
+   *     either FULL, METADATA, or MINIMAL MessageFormat
+   * @return number of unread emails from last mHours hours
+   * @throws GmailMessageFormatException if the Message objects do not contain an internalDate (and
+   *     are thus the wrong format)
+   */
+  private int findUnreadEmailCountHours(List<Message> unreadEmails)
+      throws GmailMessageFormatException {
+    // This is the oldest an email can be (in milliseconds since epoch) to be within the last mHours
+    long timeCutoffEpochMs = Instant.now().toEpochMilli() - mHours * 60 * 60 * 1000;
```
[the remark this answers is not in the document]
austinteshuba: Thanks for pointing this out!

src/main/java/com/google/sps/model/GmailResponse.java
```diff
@@ -0,0 +1,288 @@
...
+                return gmailClient.getUserMessage(message.getId(), messageFormat);
+              } catch (IOException e) {
+                throw new RuntimeException(e);
+              }
+            })
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Get the amount of unread emails from the last m hours given a list of user's unread emails from
```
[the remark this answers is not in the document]
austinteshuba: Agreed - thanks!

src/test/java/HttpServletResponseFake.java
```diff
@@ -0,0 +1,224 @@
...
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.Collection;
+import java.util.Locale;
+import javax.servlet.ServletOutputStream;
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletResponse;
+
+/** Fake version of HttpServletResponse used for testing */
+public class HttpServletResponseFake implements HttpServletResponse {
```
antmar, line 24: You could avoid all the unsupported method implementations, by using Mockito to delegate calls to your object: https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#27

- This object doesn't implement `HttpServletResponse`, and only declares the methods you want to support.
- Mockito is used to create a mock and forward all calls to your object.
  - The mock will implement `HttpServeletResponse` and all its methods.
- Methods that you have not implemented will produce some sort of error.

src/main/java/com/google/sps/model/GmailClient.java
austinteshuba: @antmar Good suggestion! I'll use String.join and when completing the finishing touches I'll look at using a builder pattern. I logged an issue for this (#100)

src/main/java/com/google/sps/model/GmailResponse.java
```diff
@@ -0,0 +1,288 @@
...
+
+  /**
+   * Create a GmailResponse instance
+   *
+   * @param nDays number of days relevant statistics are based on
+   * @param mHours number of hours relevant statistics are based on
+   * @param gmailClient GmailClient implementation with valid google credential. Will be used to
+   *     compute other statistics for GmailResponse.
+   */
+  public GmailResponse(int nDays, int mHours, GmailClient gmailClient) throws IOException {
```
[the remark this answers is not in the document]
austinteshuba: Hmm fair enough! I moved this to a utility and I'll leave the response objects simple from now on (just containing data)

src/main/java/com/google/sps/model/GmailResponse.java
austinteshuba: I actually refactored this entire function, using helper methods to make it more readable and split up the work. It's still an O(n) solution, but I think it's more readable. Let me know your thoughts!

src/main/java/com/google/sps/model/GmailResponse.java
austinteshuba: I ended up getting rid of this class. However, I used a new comparator class which I did declare at the class level inside `GmailResponseUtility`.

I completely agree with you - the readability gain is worth it, especially since the class is private.

src/main/java/com/google/sps/model/GmailResponse.java
austinteshuba: I agree with you for this case. However, now that I've moved a lot of the logic into utility methods that are standalone, I think having multiple copies of the checks is important in case one of the new utility methods is used on their own. 

Let me know what you think with the new changes

src/main/java/com/google/sps/model/GmailResponse.java
austinteshuba: Thanks for this! I ended up refactoring this method, but I'll keep this in mind for the future

src/test/java/GmailClientTest.java
austinteshuba: Fair enough - for booleans I'll just use the values.

src/test/java/GmailClientTest.java
austinteshuba: Fair enough - for booleans I'll just use the values.

src/test/java/HttpServletResponseFake.java
austinteshuba: This is a really cool tool, thanks for pointing this out! 

However, I don't know if this will work for this implementation, mainly due to `stringWriter`. If I use `response = Mockito.mock(HttpServletResponse.class, AdditionalAnswers.delgateTo(HttpServletResponseFake.class));` all methods called on that mock have to be defined in the fake. However, `HttpServletResponse` out of the box does not contain a `getStringWriter()` method, which means if I used this mock I have no way of getting the printed response. This setting also cannot be overriden (saying `Mockito.when(response.getWriter()).thenReturn(stringWriter)` does not work). 

I could just use a regular mock and do `Mockito.when(methodOne()).thenReturn(HttpServletResponseFake.methodOne())` or something along those lines, but that is going to get really messy, especially when the mocked methods take in parameters. Overall, while verbose, it might be cleaner to just create the fake.

src/test/java/HttpServletResponseFake.java
austinteshuba: Done! Thanks!

austinteshuba:
To future reviewers: Now that the utility methods are not inside of GmailResponse, do they need their own tests? If so, how should I adjust the GmailServlet tests?

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -66,15 +71,58 @@ public void doGet(
...
+      throws IOException {
+    GmailClient.MessageFormat messageFormat = GmailClient.MessageFormat.METADATA;
+
+    List<Message> unreadMessages =
+        GmailResponseUtility.getUnreadEmailsFromNDays(gmailClient, messageFormat, nDays);
+    int unreadEmailsFromNDays = unreadMessages.size();
+    int unreadEmailsFromMHours = GmailResponseUtility.countEmailsFromMHours(unreadMessages, mHours);
+    int unreadImportantEmailsFromNDays = GmailResponseUtility.countImportantEmails(unreadMessages);
+    String mostFrequentSender =
+        GmailResponseUtility.findMostFrequentSender(unreadMessages).orElse("");
```
gbuenoandrade, line 120: Can you use dependency injection to access `GmailResponseUtility`? That way you will be able to test the logic of `GmailServlet` separately (which will also make testing it very simple)

src/test/java/GmailServletTest.java
```diff
@@ -12,25 +12,18 @@
 // See the License for the specific language governing permissions and
```
gbuenoandrade, line 12: Please see my comment regarding using DI to access `GmailResponseUtility`. That should greatly simplify these tests.

src/test/java/GmailServletTest.java
austinteshuba: Just updated - sorry for missing this comment!

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -66,15 +71,58 @@ public void doGet(
...
-    response.getWriter().println(messageJson);
+    if (nDays < 0 || mHours < 0) {
+      response.sendError(400, "nDays and mHours must be positive");
+      return;
+    }
+
+    if (mHours > nDays * 24) {
+      response.sendError(400, "mHours must represent less time than nDays");
+      return;
+    }
```
antmar, line 101: Nice error handling!

src/main/java/com/google/sps/utility/GmailResponseUtility.java
```diff
@@ -0,0 +1,252 @@
...
+        .map(
+            (message) -> {
+              try {
+                return gmailClient.getUserMessage(message.getId(), messageFormat);
+              } catch (IOException e) {
+                throw new RuntimeException(e);
+              }
+            })
+        .collect(Collectors.toList());
+  }
```
antmar: I think this method belongs on `GmailClient` instead. No other method in this utility depends on the client - this one stands out.

src/main/java/com/google/sps/servlets/GmailServlet.java
antmar: I am not actually convinced this simplifies things (without sacrificing test coverage), but I look forward to the result of the refactoring.

src/main/java/com/google/sps/utility/GmailResponseUtility.java
```diff
@@ -0,0 +1,252 @@
...
+                (message) ->
+                    message.getLabelIds() != null && message.getLabelIds().contains("IMPORTANT"))
+            .count();
+  }
+
+  /**
+   * Get the sender of the most unread emails from the last nDays days in a user's Gmail account
+   *
+   * @param unreadEmails a list of unread emails (with METADATA OR FULL MessageFormat)
+   * @return sender of the most unread emails from the last nDays days. Null if no unread emails.
```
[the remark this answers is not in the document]
austinteshuba: I removed the part about the null, since that's likely clear from the use of Optional. Was there something else here that's outdated as well?

src/test/java/GmailClientTest.java
```diff
@@ -119,14 +116,30 @@ public void getQueryStringFromAnySender() {
...
-        GmailClient.emailQueryString(ONE_UNIT_OF_TIME, DAYS_UNIT, RETURN_UNREAD_ONLY, SAMPLE_EMAIL);
+        GmailClient.emailQueryString(ONE_UNIT_OF_TIME, DAYS_UNIT, true, true, SAMPLE_EMAIL);
 
-    Assert.assertTrue(multipleFilterQuery.contains(ONE_DAY_QUERY));
-    Assert.assertTrue(multipleFilterQuery.contains(UNREAD_EMAILS_QUERY));
-    Assert.assertTrue(multipleFilterQuery.contains(FROM_EMAIL_QUERY));
+    Assert.assertTrue(multipleFilterQuery.contains(ONE_DAY_QUERY.trim()));
+    Assert.assertTrue(multipleFilterQuery.contains(UNREAD_EMAILS_QUERY.trim()));
+    Assert.assertTrue(multipleFilterQuery.contains(IMPORTANT_EMAILS_QUERY.trim()));
+    Assert.assertTrue(multipleFilterQuery.contains(FROM_EMAIL_QUERY.trim()));
```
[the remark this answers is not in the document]
austinteshuba: Fixed!

src/test/java/GmailResponseUtilityTest.java
```diff
@@ -0,0 +1,94 @@
...
+    Optional<String> sender = GmailResponseUtility.findMostFrequentSender(NO_MESSAGES);
+    Assert.assertFalse(sender.isPresent());
+  }
+
+  @Test
+  public void calculateUnreadEmailsMHours() {
+    int unreadEmailsMHours =
+        GmailResponseUtility.countEmailsFromMHours(
+            SOME_MESSAGES_HALF_WITHIN_M_HOURS, DEFAULT_M_HOURS);
+    Assert.assertEquals(SOME_MESSAGES_HALF_WITHIN_M_HOURS.size() / 2, unreadEmailsMHours);
```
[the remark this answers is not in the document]
austinteshuba: Fixed! Good idea

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: Done

src/test/java/GmailServletTest.java
austinteshuba: Done

src/main/java/com/google/sps/model/GmailResponse.java
```diff
@@ -14,64 +14,31 @@
...
-  }
-
-  public int getNDays() {
-    return nDays;
-  }
-
-  public int getMHours() {
-    return mHours;
+  public GmailResponse(
+      int unreadEmailsDays, int unreadEmailsHours, int unreadImportantEmails, String sender) {
```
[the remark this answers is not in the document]
austinteshuba: I'll keep this in mind for later!

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: Thank you!

src/main/java/com/google/sps/utility/GmailResponseUtility.java
austinteshuba: Agreed - done!

src/main/java/com/google/sps/servlets/GmailServlet.java
```diff
@@ -103,21 +110,20 @@ public void doGet(
...
    * @throws IOException if an issue occurs with the Gmail Service while obtaining emails
    */
   private GmailResponse generateGmailResponse(int nDays, int mHours, GmailClient gmailClient)
       throws IOException {
     GmailClient.MessageFormat messageFormat = GmailClient.MessageFormat.METADATA;
 
-    List<Message> unreadMessages =
-        GmailResponseUtility.getUnreadEmailsFromNDays(gmailClient, messageFormat, nDays);
+    List<Message> unreadMessages = gmailClient.getUnreadEmailsFromNDays(messageFormat, nDays);
     int unreadEmailsFromNDays = unreadMessages.size();
```
antmar, line 122: Move the unread message count calculation into `GmailResponseHelper`. Otherwise, the separation of concerns looks really arbitrary. This is especially apparent when looking at the unit tests. The tests for unread e-mail count are subtly different from the tests for the other parts of the response, and it's confusing.

src/main/java/com/google/sps/utility/GmailResponseUtility.java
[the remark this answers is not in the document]
gbuenoandrade: I was only referring to the null part.

RESOLVED

src/main/java/com/google/sps/servlets/GmailServlet.java
austinteshuba: RESOLVED

PR #87, opened 2020-07-17
Refactor u-vertical-center

austinteshuba:
No visual differences here, however this refactor changes the use of `u-vertical-center` so that `&nbsp;` is not needed around spans. These generally have side effects for style (mostly in terms of line-wrapping, since they are non breaking spaces), and should be avoided when possible. 
Screenshot 2020-07-16 at 9 04 51 PM

Closes issue #85 

Dependent on PR #86


src/main/webapp/index.html
```diff
@@ -41,65 +41,65 @@ <h2 class="panel-row__header u-align-with-first-panel u-text-shadow">Services (a
       <!-- Gmail container -->
       <div class="panel u-border-mail">
         <div class="panel__header">
-          <div class="panel__content-entry u-remove-top-margin">
+          <div class="panel__content-entry u-remove-top-margin u-vertical-center">
```
tttzach, line 44: At a glance, it looks like `u-vertical-center` is always a class alongside `panel__content-entry`. Unless there is a case where `panel__content-entry` is used without `u-vertical-center`, would it be possible to merge these two together?

src/main/webapp/style/abstracts/_utilities.css
```diff
@@ -29,7 +29,8 @@
 }
 
 /*
-  Utility to vertically contents of a container
+  Utility to vertically contents of a container.
```
tttzach, line 32: extreme nit: did you mean `... vertically align contents ...`

src/main/webapp/index.html
austinteshuba: Agreed - I'm getting rid of this utility. I initially thought it would be helpful, but it's actually just redundant.

src/main/webapp/style/abstracts/_utilities.css
austinteshuba: I did, but I deleted the utility, so this is also deleted now.

src/main/webapp/index.html
```diff
@@ -31,7 +31,7 @@ <h1 class="header__heading-text u-text-shadow">Blueprint</h1>
   <div id="google-sign-in-btn"></div>
 </div>
 <!-- Contains features (gmail, tasks, calendar, etc). Hidden when user not authenticated -->
-<!-- TODO: Fix issues with u-vertical-center (Issue #85) -->
+<!-- TODO: Fix issues with (Issue #85) -->
```
antmar, line 34: Is this TODO still needed?

src/main/webapp/index.html
```diff
@@ -79,27 +79,27 @@ <h3 class="panel__section-heading u-vertical-center">Mail</h3>
             <a class="panel__content-entry-link" target="_blank" href="https://calendar.google.com/calendar/b/0/r?opentasks=1">
               <img class="panel__logo" alt="tasks-logo" src="assets/tasks.png">
             </a>
-            <h3 class="panel__section-heading u-vertical-center">Tasks</h3>
+            <h3 class="panel__section-heading">Tasks</h3>
           </div>
           <p class="panel__section-subheading">Selected Tasks:
           <div class="panel__content">
-            <div class="panel__content-entry">
+            <div class="panel__content-entry ">
```
antmar, line 86: Nit: extra space in `class` attribute.

src/main/webapp/index.html
austinteshuba: Oops - fixed

src/main/webapp/index.html
austinteshuba: Good catch! Fixed

PR #88, opened 2020-07-17
Create MVP version of Tasks Panel

tttzach:
Populates the Tasks panel to show number of tasks to complete, tasks due today, tasks completed today and tasks overdue for ALL of the user's tasks in ALL their tasklists.
TODO: Allow user to select tasklists on Tasks panel. (Issue #96)

User has 4 TaskLists
Screenshot 2020-07-21 at 12 13 07 AM

User has 2 items in Errands TaskList which have no due dates specified (just to clarify - errands are fictional) -> 2 tasks to complete
Screenshot 2020-07-21 at 12 13 21 AM

User fills car petrol in and marks `Fill car petrol` as complete in Errands TaskList, this task gets hidden automatically -> 2-1=1 task to complete, 1 completed today
Screenshot 2020-07-21 at 12 15 09 AM

User has no items in Empty TaskList
Screenshot 2020-07-21 at 12 13 29 AM

User has 3 items in Work TaskList, all of which are overdue -> 1+3=4 tasks to complete, 3 overdue
Screenshot 2020-07-21 at 12 13 40 AM

User has 2 items in Urgent TaskList, all of which are due today -> 4+2=6 tasks to complete, 2 due today
Screenshot 2020-07-21 at 12 13 50 AM

As a result, we get
Screenshot 2020-07-21 at 12 16 20 AM


src/main/java/com/google/sps/model/TasksResponse.java
```diff
@@ -0,0 +1,72 @@
...
+package com.google.sps.model;
+
+import java.util.List;
+
+/**
+ * Contains the summary tasks information that should be passed to the client, as well as the
+ * methods to generate these statistics.
+ */
+public class TasksResponse {
+  private List<String> taskListNames;
```
austinteshuba, line 24: Make these final if possible

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -52,6 +61,70 @@ public TasksServlet(
...
+    }
+    return tasks;
+  }
+
+  private List<String> getTaskListTitles(List<TaskList> taskLists) throws IOException {
+    return taskLists.stream().map(taskList -> taskList.getTitle()).collect(Collectors.toList());
+  }
+
+  private int getTasksToComplete(List<Task> tasks) {
+    List<Task> tasksCompletedToday =
```
austinteshuba, line 78: Will this return all tasks that have been completed, or just the ones completed today? If there is a restriction on the input to make the latter true, document this in a comment. This might actually be the variable name that's confusing me here, since the method is called `getTasksToComplete` but the variable is called `tasksCompletedToday`

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -52,6 +61,70 @@ public TasksServlet(
...
+  }
+
+  private int getTasksOverdue(List<Task> tasks) {
+    ZoneId zoneId = ZoneId.systemDefault();
+    String zoneOffset = zoneId.getRules().getOffset(Instant.now()).toString();
+    Instant endOfDay = LocalDate.now(zoneId).plusDays(1).atStartOfDay(zoneId).toInstant();
+    List<Task> tasksOverdue = new ArrayList<>();
+    for (Task task : tasks) {
+      if (task.getDue() != null) {
+        // Correct default timezone UTC to system's timezone
```
austinteshuba, line 116: This is good for now, but the system's timezone may not be the same as the user's timezone. We maybe should chat offline about how to handle timezones and get this working on the server once the tasks servlets are finalized a bit more

src/main/webapp/populateData.js
```diff
@@ -70,16 +70,22 @@ function populateTasks() {
...
-        // and display it on client
-        if (tasksList.length !== 0) {
-          const tasks =
-              tasksList.map((a) => a.title).reduce((a, b) => a + '\n' + b);
-          tasksContainer.innerText = tasks;
-        } else {
-          tasksContainer.innerText = 'No tasks found';
-        }
+      .then((tasksResponse) => {
+        document
```
austinteshuba, line 74: This is a really clean way of using the query selectors! I'm going to use this in my future PRs (I don't know why but I didn't think to chain them like this. Thanks!)

src/main/webapp/populateData.js
```diff
@@ -89,6 +95,7 @@ function populateTasks() {
       });
 }
 
+
```
austinteshuba, line 98: Only one blank line is needed between methods

src/main/java/com/google/sps/model/TasksResponse.java
tttzach: I assume you meant the class, thanks for pointing that out!

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: I missed that variable, must have copied the code from another method and forgot to change the variable name. Thanks for pointing it out!

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: Agreed!

src/main/webapp/populateData.js
tttzach: Thanks! Honestly, it was to shush the JavaScript linter.

src/main/webapp/populateData.js
tttzach: Good catch, thanks!

src/main/java/com/google/sps/model/TasksResponse.java
tttzach: Not sure why I didn't do this in the first place, thanks!

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -52,6 +61,71 @@ public TasksServlet(
     this.tasksClientFactory = tasksClientFactory;
   }
 
+  private List<Task> getTasks(TasksClient tasksClient) throws IOException {
+    List<TaskList> taskLists = tasksClient.listTaskLists();
+    List<Task> tasks = new ArrayList<>();
+    for (TaskList taskList : taskLists) {
+      tasks.addAll(tasksClient.listTasks(taskList));
+    }
```
gbuenoandrade, line 69: Alternative: `taskLists.forEach(l -> tasks.addAll(taskClient.listTasks(l)))` (feel free to choose)

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -52,6 +61,71 @@ public TasksServlet(
...
+      }
+    }
+    return tasksCompletedToday.size();
+  }
+
+  private int getTasksOverdue(List<Task> tasks) {
+    ZoneId zoneId = ZoneId.systemDefault();
+    String zoneOffset = zoneId.getRules().getOffset(Instant.now()).toString();
+    Instant endOfDay = LocalDate.now(zoneId).plusDays(1).atStartOfDay(zoneId).toInstant();
+    List<Task> tasksOverdue = new ArrayList<>();
```
gbuenoandrade, line 114: If you only care about the size, maybe you should just have a counter?

src/test/java/TasksServletTest.java
```diff
@@ -112,48 +143,109 @@ public void noTaskLists() throws IOException, ServletException {
     Mockito.when(tasksClient.listTaskLists()).thenReturn(NO_TASKLISTS);
     servlet.doGet(request, response);
```
gbuenoandrade, line 144: Please use Austin's fake response when https://github.com/googleinterns/step7-2020/pull/86 lands

src/main/java/com/google/sps/model/TasksClientImpl.java
```diff
@@ -46,7 +46,8 @@ private TasksClientImpl(Credential credential) {
   @Override
   public List<Task> listTasks(TaskList taskList) throws IOException {
     // returns null if no tasks exist. Convert to empty list for ease.
-    List<Task> tasks = tasksService.tasks().list(taskList.getId()).execute().getItems();
+    List<Task> tasks =
+        tasksService.tasks().list(taskList.getId()).setShowHidden(true).execute().getItems();
```
antmar, line 50: Please update the documentation of the `TasksClient` interface to indicate that hidden/completed tasks are being returned as well. That's not immediately obvious. Also add a comment here explaining what `setShowHidden` does.

src/main/java/com/google/sps/model/TasksResponse.java
```diff
@@ -0,0 +1,72 @@
...
+   * @param tasksDueToday number of tasks the user has yet to complete for today.
+   * @param tasksCompletedToday number of tasks the user has marked as complete for today.
+   * @param tasksOverdue number of tasks that have due dates before today.
+   */
+  public TasksResponse(
+      List<String> taskListNames,
+      int tasksToComplete,
+      int tasksDueToday,
+      int tasksCompletedToday,
+      int tasksOverdue) {
```
antmar, line 45: This constructor is just at the edge of being too complicated. If you were to add any more fields, I would recommend switching to a Builder pattern for initializing this object.

The concern here is that it would be very easy to get the order of the parameters wrong, and there is no way to see that at the call site. Unit tests should help catch bugs, but it's better when the code is self-documenting.

(You don't need to change it in this PR.)

src/test/java/TasksServletTest.java
```diff
@@ -112,48 +143,109 @@ public void noTaskLists() throws IOException, ServletException {
...
     printWriter.flush();
-    List<Task> tasks = gson.fromJson(stringWriter.toString(), LIST_OF_TASKS_TYPE);
-    Assert.assertTrue(tasks.isEmpty());
+    TasksResponse tasksResponse = gson.fromJson(stringWriter.toString(), TasksResponse.class);
+    Assert.assertEquals(NO_TASKLISTS_TITLES, tasksResponse.getTaskListNames());
+    Assert.assertEquals(0, tasksResponse.getTasksToComplete());
+    Assert.assertEquals(0, tasksResponse.getTasksDueToday());
+    Assert.assertEquals(0, tasksResponse.getTasksCompletedToday());
+    Assert.assertEquals(0, tasksResponse.getTasksOverdue());
```
antmar, line 151: You could override `Object.equals` in `TasksResponse`, and then use `Assert.assertEquals` to compare `TasksResponse` objects directly.

src/test/java/TasksServletTest.java
```diff
@@ -112,48 +143,109 @@ public void noTaskLists() throws IOException, ServletException {
...
+    Mockito.when(tasksClient.listTaskLists()).thenReturn(ONE_TASKLIST);
+    Mockito.when(tasksClient.listTasks(TASKLIST_ONE)).thenReturn(TASKS_DUE_YESTERDAY);
+    servlet.doGet(request, response);
+    printWriter.flush();
+    TasksResponse tasksResponse = gson.fromJson(stringWriter.toString(), TasksResponse.class);
+    Assert.assertEquals(ONE_TASKLIST_TITLES, tasksResponse.getTaskListNames());
+    Assert.assertEquals(1, tasksResponse.getTasksToComplete());
+    Assert.assertEquals(0, tasksResponse.getTasksDueToday());
+    Assert.assertEquals(0, tasksResponse.getTasksCompletedToday());
+    Assert.assertEquals(1, tasksResponse.getTasksOverdue());
```
antmar, line 179: It's not immediately obvious that these responses are correct, just by looking at this test method. I see that you have incomplete and overdue tasks, but how do I know that `TASKS_DUE_YESTERDAY` only returns one overdue task?

If you follow my suggestion above (to allow comparing responses), then you can define static constants for the expected responses to each set of input tasks. This would put the definition of the responses closer to where the inputs are defined.

(I am not sure what you would do with more complex tests that use multiple task lists - probably just define responses for those combinations.)

src/test/java/TasksServletTest.java
```diff
@@ -112,48 +143,109 @@ public void noTaskLists() throws IOException, ServletException {
...
-            TASKS_THREE_FOUR.get(0),
-            TASKS_THREE_FOUR.get(1)));
+    TasksResponse tasksResponse = gson.fromJson(stringWriter.toString(), TasksResponse.class);
+    Assert.assertEquals(ONE_TASKLIST_TITLES, tasksResponse.getTaskListNames());
+    Assert.assertEquals(3, tasksResponse.getTasksToComplete());
+    Assert.assertEquals(1, tasksResponse.getTasksDueToday());
+    Assert.assertEquals(1, tasksResponse.getTasksCompletedToday());
+    Assert.assertEquals(1, tasksResponse.getTasksOverdue());
   }
 }
```
antmar, line 251: You should add a test with more than one task list.

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: The alternative is great but I can't seem to handle the IOException from calling `tasksClient.listTasks(l)` inside `.addAll`. I would have to surround it in a try catch which I think is more clutter than the for loop. Any thoughts?
```
    taskLists.forEach(taskList -> {
      try {
        tasks.addAll(tasksClient.listTasks(taskList));
      } catch (IOException e) {
        // TODO Auto-generated catch block
        e.printStackTrace();
      }
    });
```

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: Agreed! I have done the same for `getTasksCompletedToday`. Should I do the same for `getTasksToComplete` and `getTasksDueToday`? This would entail using an explicit for loop instead of streams. I imagine that it would be better to use an explicit for loop + save some memory rather than using streams but I could be wrong.

src/test/java/TasksServletTest.java
tttzach: Ok sure!

src/main/java/com/google/sps/model/TasksResponse.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: `RESOLVED`

src/main/webapp/populateData.js
tttzach: `RESOLVED`

src/main/webapp/populateData.js
tttzach: `RESOLVED`

antmar:
You are welcome to address my comments about the tests in a follow-up PR. Do add the code comments, however.

src/main/java/com/google/sps/model/TasksResponse.java
tttzach: I agree, will definitely look into refactoring this using the Java annotation system you [linked](https://projectlombok.org/features/Builder) in our group chat.

src/main/java/com/google/sps/model/TasksClientImpl.java
tttzach: Fair enough, thanks!

src/test/java/TasksServletTest.java
tttzach: This would make it a lot neater, thanks!

src/test/java/TasksServletTest.java
tttzach: Added, thanks!

src/main/java/com/google/sps/servlets/TasksServlet.java
gbuenoandrade: Oh, Java handling checked exceptions weirdly. You did well sticking with the for loop, then

src/main/java/com/google/sps/model/TasksClientImpl.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/TasksServlet.java
gbuenoandrade: Maybe you could still use [.count()](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#count--)?

src/test/java/TasksServletTest.java
tttzach: `RESOLVED` https://github.com/googleinterns/step7-2020/pull/105

src/test/java/TasksServletTest.java
tttzach: `RESOLVED` https://github.com/googleinterns/step7-2020/pull/105

src/test/java/TasksServletTest.java
tttzach: `RESOLVED` https://github.com/googleinterns/step7-2020/pull/105

src/main/java/com/google/sps/model/TasksResponse.java
tttzach: `RESOLVED` https://github.com/googleinterns/step7-2020/pull/105

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: Didn't know about `.count()` - it's perfect, thanks!

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: Hahah cool!

tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: `RESOLVED` #105

src/test/java/TasksServletTest.java
tttzach: `RESOLVED` #105

tttzach:
All review comments here have been resolved either in this PR or #105 if it says so after `RESOLVED!

PR #91, opened 2020-07-17
Implement JSON utility class (Remerge of #40)

tttzach:
This is a rebase and remerge for #40 which was accidentally merged into feature/calendar-utility.


tttzach:
Thanks Anton

PR #98, opened 2020-07-21
POST requests for tasks

austinteshuba:
Allows for adding new tasks and tasklists to the server, as well as allows for the client to get a list of all tasklists and their associated tasks. 

This will be required for the assign panel (and other panels that use tasks).


austinteshuba:
@austinteshuba Clean up before assigning

austinteshuba:
Split into different PRs #102 #103

PR #99, opened 2020-07-21
Initial implementation of the calendar panel for the MVP.

AymarMuhikira:
Display the free hours in the calendar panel, both work and personal time.
Screenshot 2020-07-21 at 4 53 14 PM - Display 2
The event duration is subtracted from either the personal or work time depending on the beginning of the event.
Will later work on having the possibility to differentiate between the work hours and personal time for a single event that overlaps both.
The work hours are defined as being between 10:00 am to 6:00 pm. The time is given as a long representing the Unix time. The timezone used is UTC.
The code to calculate the free time left is not very accurate and fails to account for many edge cases, will have to refactor the method as stated in Issue #104


src/main/java/com/google/sps/data/CalendarClientData.java
```diff
@@ -0,0 +1,77 @@
+// Copyright 2019 Google LLC
```
tttzach, line 1: extreme nit: 2020

src/main/java/com/google/sps/data/CalendarClientData.java
```diff
@@ -0,0 +1,77 @@
...
+package com.google.sps.data;
+
+import java.util.*;
+
+/** Class containing the user's free hours. */
+public final class CalendarClientData {
+
+  private final int startDay;
+  private final int[] workHours;
+  private final int[] personalHours;
```
tttzach, line 24: nit: Communicating a [review comment](https://github.com/googleinterns/step7-2020/pull/63#discussion_r455246251) I received from Gui here, I think `List<int>` would be preferable even if that includes conversion.

src/main/java/com/google/sps/data/CalendarClientData.java
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Initialize the class by calculating the start day. The free hours should be eight hours for
+   * the working hours and sixteen hours for the pesonal time initially. 
+   *
+   * @param startTime parameter of type long that gives the Unix epoch time of the start/now. The timezone is UTC
+   */
+  public CalendarClientData(long startTime) {
+    long numMillisecondsDay = 24 * 60 * 60 * 1000;
```
tttzach, line 34: Might be worth changing these variables to constants since it seems like you use them in `addEvent` as well.

src/main/webapp/index.html
```diff
@@ -119,38 +119,38 @@ <h3 class="panel__section-heading u-vertical-center">Calendar</h3>
           <p class="panel__section-subheading">Selected Calendars:
           <div class="panel__content">
             <div class="panel__content-entry">
-              <p class="panel__text-icon u-text-calendar">M:
+              <p class="panel__text-icon u-text-calendar" id="day-0">M:
               <div>
-                <p class="u-block">3h 4m free (working)
-                <p class="u-block">4h 12m free (personal)
+                <p class="u-block" id="work-0">3h 4m free (working)
+                <p class="u-block" id="personal-0">4h 12m free (personal)
```
tttzach, line 125: extreme nit: maybe `work-day-0` to make it extremely clear?

src/main/webapp/populateData.js
```diff
@@ -102,17 +102,18 @@ function populateCalendar() {
...
-        // Convert JSON to string containing all event summaries
-        // and display it on client
-        // Handle case where user has no events to avoid unwanted behaviour
-        if (eventList.length !== 0) {
-          const events =
-              eventList.map((a) => a.summary).reduce((a, b) => a + '\n' + b);
-          calendarContainer.innerText = events;
-        } else {
-          calendarContainer.innerText = 'No events in the calendar';
+      .then((hoursJson) => {
```
tttzach, line 105: `mvn spotless:apply` doesn't deal with linting JavaScript so make sure you manually lint according to the suggestions

src/main/webapp/populateData.js
```diff
@@ -102,17 +102,18 @@ function populateCalendar() {
...
+        // Display the days and the free hours for each one of them
+        let days = ['Mo', 'Tu', 'We', 'Th', 'Fr', 'Sa', 'Su'];
+        for (var day = 0; day < 5; day++) {
+          const dayContainer = document.querySelector('#day-'+day);
+          dayContainer.innerText = days[hoursJson.startDay + day];
+          const workContainer = document.querySelector('#work-'+day);
+          workContainer.innerText = convertTime(hoursJson.workHours[day]) + ' (working)';
+          const personalContainer = document.querySelector('#personal-'+day);
+          personalContainer.innerText = convertTime(hoursJson.personalHours[day]) + ' (personal)';
```
tttzach: Do you use these const variables?

src/test/java/CalendarServletTest.java
```diff
@@ -117,48 +122,40 @@ public void setUp() throws IOException, GeneralSecurityException {
...
-  public List<Event> checkReturn() throws IOException, ServletException {
-    // Method that handles the once the Calendar Client has been mocked
+  private CalendarClientData getServletResponse() throws IOException, ServletException {
+    // Method that handles the request once the Calendar Client has been mocked
     servlet.doGet(request, response);
     printWriter.flush();
     String actualString = stringWriter.toString();
-    Type type = new TypeToken<List<Event>>() {}.getType();
-    List<Event> actual = gson.fromJson(actualString, type);
+    CalendarClientData actual = gson.fromJson(actualString, CalendarClientData.class);
```
tttzach, line 159: Nice refactor!

src/main/webapp/index.html
```diff
@@ -119,38 +119,38 @@ <h3 class="panel__section-heading u-vertical-center">Calendar</h3>
...
               </div>
             </div>
             <div class="panel__content-entry">
-              <p class="panel__text-icon u-text-calendar">T:
+              <p class="panel__text-icon u-text-calendar" id="day-1">T:
               <div>
-                <p class="u-block">3h 4m free (working)
-                <p class="u-block">4h 12m free (personal)
+                <p class="u-block" id="work-1">3h 4m free (working)
+                <p class="u-block" id="personal-1">4h 12m free (personal)
```
tttzach, line 132: You can remove these hardcoded values now that your JavaScript function populates them.

src/test/java/CalendarServletTest.java
```diff
@@ -75,22 +79,23 @@
...
+  private static final DateTime CURRENT_TIME = new DateTime("2020-05-19T09:00:00+00:00");
+  private static final DateTime END_TIME = new DateTime(CURRENT_TIME.getValue() + 5 * 24 * 60 * 60 * 1000);
+  private static final DateTime EVENT_ONE_START = new DateTime("2020-05-19T15:00:00+00:00");
+  private static final DateTime EVENT_ONE_END = new DateTime("2020-05-19T16:00:00+00:00");
+  private static final DateTime EVENT_TWO_START = new DateTime("2020-05-20T06:00:00+00:00");
+  private static final DateTime EVENT_TWO_END = new DateTime("2020-05-20T08:00:00+00:00");
+  private static final EventDateTime startOne = new EventDateTime().setDateTime(EVENT_ONE_START);
+  private static final EventDateTime endOne = new EventDateTime().setDateTime(EVENT_ONE_END);
+  private static final EventDateTime startTwo = new EventDateTime().setDateTime(EVENT_TWO_START);
+  private static final EventDateTime endTwo = new EventDateTime().setDateTime(EVENT_TWO_END);
```
tttzach, line 90: I could be wrong - would this be in constant case as well?

src/test/java/CalendarServletTest.java
```diff
@@ -117,48 +122,40 @@ public void setUp() throws IOException, GeneralSecurityException {
   public void noCalendarEvent() throws IOException, ServletException {
     // Test case where there are no events in the user's calendar
     Mockito.when(calendarClient.getCalendarList()).thenReturn(ONE_CALENDAR);
-    Mockito.when(calendarClient.getCalendarEvents(PRIMARY)).thenReturn(NO_EVENT);
-    List<Event> actual = checkReturn();
-    Assert.assertEquals(NO_EVENT, actual);
+    Mockito.when(calendarClient.getUpcomingEvents(PRIMARY, CURRENT_TIME, END_TIME)).thenReturn(NO_EVENT);
+    Mockito.when(calendarClient.getCurrentTime()).thenReturn(CURRENT_TIME);
+    CalendarClientData actual = getServletResponse();
+    int hour = 60 * 60 * 1000;
```
tttzach, line 127: nit: Maybe make this a constant, it is used in multiple tests.

src/main/java/com/google/sps/data/CalendarClientData.java
AymarMuhikira: Thank you. `Resolved`

src/main/java/com/google/sps/data/CalendarClientData.java
AymarMuhikira: Changed to using `List`. `Resolved`

src/main/java/com/google/sps/data/CalendarClientData.java
AymarMuhikira: Good idea, I made this a global variable for the class.

src/main/webapp/index.html
AymarMuhikira: I changed to `work-day-0` to make this clear. I can not use the day names in the id because I am displaying 5 days (not the entire week), so the day names changed depending on the current day.

src/main/webapp/populateData.js
AymarMuhikira: I am using them to set the inner text.

src/test/java/CalendarServletTest.java
AymarMuhikira: Thank you!

src/main/webapp/index.html
AymarMuhikira: Good suggestion. `Resolved`

src/test/java/CalendarServletTest.java
AymarMuhikira: Agreed, changed this to a global variable. `Resolved`

src/main/java/com/google/sps/data/CalendarClientData.java
```diff
@@ -0,0 +1,86 @@
...
+      personalHours.set(
+          beginDay - startDay, Math.max(personalHours.get(beginDay - startDay) - duration, 0));
+    } else {
+      workHours.set(
+          beginDay - startDay, Math.max(workHours.get(beginDay - startDay) - duration, 0));
+    }
+  }
+
+  public int getStartDay() {
+    // Usefull for testing purposes
```
tttzach, line 73: typo nit: Did you mean useful?

src/main/webapp/populateData.js
```diff
@@ -102,17 +102,21 @@ function populateCalendar() {
...
+          dayContainer.innerText = days[hoursJson.startDay + day];
+          const workContainer = document.querySelector('#work-day-'+day);
+          workContainer.innerText =
+              convertTime(hoursJson.workHours[day]) + ' (working)';
+          const personalContainer =
+              document.querySelector('#personal-day-'+day);
+          personalContainer.innerText =
+              convertTime(hoursJson.personalHours[day]) + ' (personal)';
         }
+        console.log(hoursJson);
```
tttzach, line 119: nit: was this for debugging?

src/main/java/com/google/sps/data/CalendarClientData.java
AymarMuhikira: `Resolved`

src/main/webapp/populateData.js
AymarMuhikira: Yes, it was. I removed it.

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -67,14 +69,46 @@ public void doGet(
         : "Null credentials (i.e. unauthenticated requests) should already be handled";
 
     CalendarClient calendarClient = calendarClientFactory.getCalendarClient(googleCredential);
-    List<Event> calendarEvents = getEvents(calendarClient);
+    long fiveDays = 5 * 24 * 60 * 60 * 1000;
```
gbuenoandrade, line 72: Nit: `TimeUnit` lib

src/main/java/com/google/sps/model/CalendarClientImpl.java
```diff
@@ -60,6 +61,29 @@ private CalendarClientImpl(Credential credential) {
     return events != null ? events : new ArrayList<>();
   }
 
+  @Override
+  public List<Event> getUpcomingEvents(
+      CalendarListEntry calendarList, DateTime timeMin, DateTime timeMax) throws IOException {
```
gbuenoandrade, line 66: Suggestion: Have you taken a look at [Date](https://docs.oracle.com/javase/8/docs/api/java/util/Date.html)? I think that might be slightly more convenient to work with. Particularly, that allows you to compare them using methods like `before()` and `after()`.

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -67,14 +69,46 @@ public void doGet(
...
+    }
 
     // Convert event list to JSON and print to response
     Gson gson = new Gson();
-    String eventJson = gson.toJson(calendarEvents);
+    String hoursJson = gson.toJson(calendarClientData);
 
     response.setContentType("application/json");
-    response.getWriter().println(eventJson);
+    response.getWriter().println(hoursJson);
```
antmar, line 92: Use the `JsonUtil`

src/test/java/CalendarServletTest.java
```diff
@@ -114,49 +120,47 @@ public void setUp() throws IOException, GeneralSecurityException {
...
+    Assert.assertTrue(personalHours.equals(actual.getPersonalHours()));
   }
 
-  private List<Event> getServletResponse() throws IOException, ServletException {
-    // Method that handles the once the Calendar Client has been mocked
+  private CalendarClientData getServletResponse() throws IOException, ServletException {
+    // Method that handles the request once the Calendar Client has been mocked
     servlet.doGet(request, response);
     printWriter.flush();
     String actualString = stringWriter.toString();
```
antmar, line 162: Use the `HttpServletResponseFake`

src/main/java/com/google/sps/data/CalendarClientData.java
```diff
@@ -0,0 +1,87 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.util.*;
+
+/** Class containing the user's free hours. */
+public final class CalendarClientData {
```
antmar, line 20: This class is directly serialized into the response using GSON, which serializes all private fields. The *private* field types/names of this class are part of the *public* API of your servlet. Therefore, I think it should be called `CalendarDataResponse` and contain only fields, constructor, and getters.

Move the logic into a separate utility. (Keep the initialize + addEvent pattern - I like it!) You can either have a method on the utility called getCalendarDataResponse, or just create the response in your servlet.

src/main/java/com/google/sps/data/CalendarClientData.java
```diff
@@ -0,0 +1,87 @@
...
+
+  private final int startDay;
+  private final List<Integer> workHours;
+  private final List<Integer> personalHours;
+  private final long numMillisecondsDay = 24 * 60 * 60 * 1000;
+  private final int Hour = 60 * 60 * 1000;
```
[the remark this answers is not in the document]
AymarMuhikira: Thank you for the information. `Resolved`

src/main/java/com/google/sps/data/CalendarClientData.java
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * Initialize the class by calculating the start day. The free hours should be eight hours for the
+   * working hours and sixteen hours for the pesonal time initially.
+   *
+   * @param startTime parameter of type long that gives the Unix epoch time of the start/now. The
+   *     timezone is UTC
+   */
+  public CalendarClientData(long startTime) {
+    int numDays = (int) Math.floor(startTime / numMillisecondsDay);
```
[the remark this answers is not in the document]
AymarMuhikira: I converted to using long. `Resolved`

src/main/java/com/google/sps/data/CalendarClientData.java
```diff
@@ -0,0 +1,87 @@
...
+    this.startDay = ((numDays + 3) % 7);
```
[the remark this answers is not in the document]
AymarMuhikira: Agreed, I changed it to a constant. `Resolved`

src/main/java/com/google/sps/data/CalendarClientData.java
```diff
@@ -0,0 +1,87 @@
...
+    int eightHours = 8 * Hour;
+    int sixteenHours = 16 * Hour;
+    this.workHours =
+        new ArrayList<>(Arrays.asList(eightHours, eightHours, eightHours, eightHours, eightHours));
```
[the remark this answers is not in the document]
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/data/CalendarClientData.java
```diff
@@ -0,0 +1,87 @@
...
+   *
+   * @param startTime parameter of type long that gives the Unix epoch time of the start of the
+   *     event
+   * @param endTime parameter of type long that gives the Unix epoch time of the end of the event
+   */
+  public void addEvent(long startTime, long endTime) {
+    int numDays = (int) Math.floor(startTime / numMillisecondsDay);
+    int beginDay = ((numDays + 3) % 7);
+    int duration = (int) (endTime - startTime);
+    int startHour = (int) (startTime % numMillisecondsDay);
```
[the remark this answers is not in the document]
AymarMuhikira: Agreed. I changed this. `Resolved`.

src/main/java/com/google/sps/data/CalendarClientData.java
```diff
@@ -0,0 +1,87 @@
...
+    int workBegin = 10 * Hour;
+    int workEnd = 18 * Hour;
+    if (startHour < workBegin || startHour > workEnd) {
```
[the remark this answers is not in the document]
AymarMuhikira: I changed the function to account for this. `Resolved`

src/main/java/com/google/sps/data/CalendarClientData.java
```diff
@@ -0,0 +1,87 @@
...
+package com.google.sps.data;
+
+import java.util.*;
+
+/** Class containing the user's free hours. */
+public final class CalendarClientData {
+
+  private final int startDay;
+  private final List<Integer> workHours;
+  private final List<Integer> personalHours;
```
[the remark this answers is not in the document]
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/CalendarServlet.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/model/CalendarClientImpl.java
AymarMuhikira: The calendar API gives DateTime types and requires DateTime types to get and set Events.

src/main/java/com/google/sps/model/CalendarClientImpl.java
```diff
@@ -60,6 +61,29 @@ private CalendarClientImpl(Credential credential) {
...
+            .execute()
+            .getItems();
+
+    return events != null ? events : new ArrayList<>();
+  }
+
+  @Override
+  public DateTime getCurrentTime() throws IOException {
+    DateTime now = new DateTime(System.currentTimeMillis());
+    return now;
```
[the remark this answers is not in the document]
AymarMuhikira: Agreed. `Resolved`

src/main/java/com/google/sps/servlets/CalendarServlet.java
AymarMuhikira: `Resolved`

src/test/java/CalendarServletTest.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/data/CalendarClientData.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,174 @@
...
+// limitations under the License.
+
+package com.google.sps.utility;
+
+import com.google.sps.data.CalendarDataResponse;
+import java.util.*;
+import java.util.concurrent.TimeUnit;
+
+/** Class containing the user's free hours. */
+public final class FreeTimeUtility {
```
gbuenoandrade, line 22: Food for thought: I wonder if we could simplify this a bit, particularly share some logic among the different cases.
One way I thought of:
1) You could have a function `int calcIntersection(Interval event, Interval window)` that would return the intersection in hours between event and window. Then, for each event `e`, you would call `calcIntersection(e, personalWindow)` and `calcIntersection(e, workWindow)`.
2) In order to take into account the fact that events can have intersections among themselves, you could augment `calcIntersection` and make it return an intersection interval `[a, b]` and memorize those. After you are done processing all the events, you would look into those intervals and calculate the remaining free time.

Those are just ideas, you don't need to strictly follow them, but I think it's worth spending some time trying to use something along those lines to make this class a bit simpler. Please let me know if you need any clarification/help.

src/main/webapp/index.html
```diff
@@ -117,38 +117,38 @@ <h3 class="panel__section-heading">Calendar</h3>
           <p class="panel__section-subheading">Selected Calendars:
           <div class="panel__content">
             <div class="panel__content-entry">
-              <p class="panel__text-icon u-text-calendar">M:
+              <p class="panel__text-icon u-text-calendar" id="day-0">
```
antmar, line 120: Instead of hardcoding a "content-entry" for every day of the work week, you should generate these in the Javascript. That way, both the number of work days and the start of the work week can be dictated entirely by the server response.

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -69,27 +70,23 @@ public void doGet(
         : "Null credentials (i.e. unauthenticated requests) should already be handled";
 
     CalendarClient calendarClient = calendarClientFactory.getCalendarClient(googleCredential);
-    long fiveDays = 5 * 24 * 60 * 60 * 1000;
+    long fiveDays = TimeUnit.DAYS.toMillis(5);
```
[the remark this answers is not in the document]
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,174 @@
...
+  private final long startDay;
+  private final List<Long> workHoursPerDay;
+  private final List<Long> personalHoursPerDay;
+  private final long numMillisecondsDay = TimeUnit.HOURS.toMillis(24);
+  private final long Hour = TimeUnit.HOURS.toMillis(1);
+  private final int UnixEpochDayShift = 3;
+  private final long personalBegin = 7 * Hour;
+  private final long workBegin = 10 * Hour;
+  private final long workEnd = 18 * Hour;
+  private final long personalEnd = 23 * Hour;
```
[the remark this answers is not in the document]
AymarMuhikira: Changed the variables already assigned to constants. The list of constants might be different due to refactoring.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,174 @@
...
+
+  /**
+   * Initialize the class by calculating the start day. The free hours should be eight hours for the
+   * working hours and for the pesonal time initially. Adjust the free hours on the first day to
+   * account for the fact that there are hours that have already passed
+   *
+   * @param startTime parameter of type long that gives the Unix epoch time of the start/now. The
```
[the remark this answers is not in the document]
AymarMuhikira: Ok. Thank you. `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,174 @@
...
+   */
+  public void addEvent(long startTime, long endTime) {
+    long numDaysBegin = startTime / numMillisecondsDay;
+    long numDaysEnd = endTime / numMillisecondsDay;
+    long beginDay = (numDaysBegin + UnixEpochDayShift) % 7;
+    long endDay = (numDaysEnd + UnixEpochDayShift) % 7;
+    long startHour = startTime % numMillisecondsDay;
+    long endHour = endTime % numMillisecondsDay;
+    int indexBegin = (int) (beginDay - startDay);
+    int indexEnd = (int) (endDay - startDay);
```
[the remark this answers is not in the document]
AymarMuhikira: I realized this was a mistake on my part. I already make sure the event can't go beyond 5 days on the CalendarServlet, so this is a redundant check. `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  private void setAllDayEvent(int index) {
+    personalHoursPerDay.set(index, (long) 0);
+    workHoursPerDay.set(index, (long) 0);
+  }
+
+  private void setStartDayEvent(int index, long startHour) {
+    if (startHour < personalBegin) {
+      // This case is similar as an all day event
```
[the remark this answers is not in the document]
AymarMuhikira: This was because the hours allocated for this case would be beginning at the start of free time to the end of the day. The refactoring makes this a deprecated function however.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,174 @@
...
+          index, Math.max(personalHoursPerDay.get(index) - personalDuration, 0));
+    } else if (endHour >= personalBegin) {
+      // Only part of the morning is occupied
+      long personalDuration = endHour - personalBegin;
+      personalHoursPerDay.set(
+          index, Math.max(personalHoursPerDay.get(index) - personalDuration, 0));
+    }
+  }
+
+  public CalendarDataResponse getCalendarDataResponse() {
```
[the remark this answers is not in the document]
AymarMuhikira: It is being called on the servlet to get the response object that does not include methods other than constructor and getters.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
AymarMuhikira: I used a similar idea but rather calculated directly the remaining free interval based on the intervals for each event. The core method is similar to the one used in the fifth walkthrough.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,174 @@
...
+   * Initialize the class by calculating the start day. The free hours should be eight hours for the
+   * working hours and for the pesonal time initially. Adjust the free hours on the first day to
+   * account for the fact that there are hours that have already passed
+   *
+   * @param startTime parameter of type long that gives the Unix epoch time of the start/now. The
+   *     timezone is UTC
+   */
+  public FreeTimeUtility(long startTime) {
+    long numDays = startTime / numMillisecondsDay;
+    this.startDay = (numDays + UnixEpochDayShift) % 7;
```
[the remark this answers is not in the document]
AymarMuhikira: Agreed. I defined the offsets I have. `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,174 @@
...
+    long eightHours = 8 * Hour;
```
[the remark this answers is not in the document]
AymarMuhikira: `Resolved`

src/main/webapp/index.html
AymarMuhikira: I agree that having one source for the data, and not hardcoding things would be a good way to avoid errors due to un-synchronized number of days. However, having clearly defined classes for these entries allows easy css formatting, that would not be as obvious if the entire panel was set from the JS script.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,142 @@
...
+import com.google.sps.data.CalendarDataResponse;
+import java.util.*;
+import java.util.concurrent.TimeUnit;
+
+/** Class containing the user's free hours. */
+public final class FreeTimeUtility {
+
+  private final Date startDate;
+  private List<Long> workHoursPerDay;
+  private List<Long> personalHoursPerDay;
```
antmar, line 26: These lists can just be local variables in `getCalendarDataResponse`. You are not updating them in `addEvent`, and you are not using them for any other reason except to get the response.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -21,154 +21,122 @@
...
+  public FreeTimeUtility(Date startDate) {
+    this.startDate = startDate;
+    Date basisDate = new Date(startDate.getYear(), startDate.getMonth(), startDate.getDate(), 0, 0);
+
+    this.morningFreeInterval = new ArrayList<>();
+    this.workFreeInterval = new ArrayList<>();
+    this.eveningFreeInterval = new ArrayList<>();
+
+    for (int day = 0; day < 5; day++) {
+      Date workStart = new Date(basisDate.getTime() + day * DAY_MILLI + WORK_BEGIN);
```
gbuenoandrade, line 53: Nit: [Instant](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html) could make this slightly simpler. Example: https://github.com/googleinterns/step7-2020/blob/72762882bf046a75fd8c8ff2c6a7f03844277c15/src/main/java/com/google/sps/servlets/TasksServlet.java#L192

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,142 @@
...
+    return hoursPerDay;
+  }
+
+  /**
+   * Method to create and return the calendar data response.
+   *
+   * @return the calendar data response to be sent as servlet response
+   */
+  public CalendarDataResponse getCalendarDataResponse() {
+    updateFreeTime();
```
antmar, line 138: I think it would be cleaner to inline the `updateFreeTime` logic here. It's only used in this one place, and it's what produces the data you return, so having it here makes sense.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,142 @@
...
+      }
+    }
+    return updatedTime;
+  }
+
+  /** Method to calculate the free time for each day based on the free intervals. */
+  private void updateFreeTime() {
+    long zero = 0;
+    this.workHoursPerDay = Arrays.asList(zero, zero, zero, zero, zero);
+    this.personalHoursPerDay = Arrays.asList(zero, zero, zero, zero, zero);
```
antmar, line 105: You shouldn't hard-code the number of days in a week, especially when it's encoded "structurally" in a variable-length method call. Use a constant or instance variable to abstract this away.

You can initialize an array of dynamic length like this: create an array of the desired length and call `Arrays.fill`. (Also, I am pretty sure that Java initializes an array of integers to zero by default.)

(Also, making a variable for `0` is not necessary - just use the integer constant.)

src/main/java/com/google/sps/utility/FreeTimeUtility.java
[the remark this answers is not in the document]
gbuenoandrade: All right, thanks for clarifying.
RESOLVED

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,142 @@
...
+    this.workHoursPerDay = updateHoursPerDay(this.workHoursPerDay, this.workFreeInterval);
+    this.personalHoursPerDay =
+        updateHoursPerDay(this.personalHoursPerDay, this.morningFreeInterval);
+    this.personalHoursPerDay =
+        updateHoursPerDay(this.personalHoursPerDay, this.eveningFreeInterval);
+  }
+
+  /**
+   * Method to update the free time on a particular list of free time based on a particular list of
+   * intervals
```
antmar, line 115: Be more specific about what `update` means. You are accumulating (adding) hours to each day, based on the given free intervals.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,142 @@
...
+   *
+   * @param hoursPerDay the list of free time in milliseconds to update.
+   * @param freeInterval the list of free intervals
+   * @return the updated list of free time
+   */
+  private List<Long> updateHoursPerDay(List<Long> hoursPerDay, List<DatePair> freeInterval) {
+    for (DatePair interval : freeInterval) {
+      int index = interval.getKey().getDay() - this.startDate.getDay();
+      index = index < 0 ? index + OFFSET : index;
```
antmar, line 124: I am confused by what this logic does, especially when I don't know what `OFFSET` means. Please add a code comment here, and use a better name for `OFFSET` as I suggested above.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,142 @@
...
+
+/** Class containing the user's free hours. */
+public final class FreeTimeUtility {
+
+  private final Date startDate;
+  private List<Long> workHoursPerDay;
+  private List<Long> personalHoursPerDay;
+  private List<DatePair> morningFreeInterval;
+  private List<DatePair> workFreeInterval;
+  private List<DatePair> eveningFreeInterval;
```
antmar, line 29: Please add a comment here explaining the relationship between `startDate` and these lists.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,142 @@
...
+   * @param eventEnd parameter that gives end tome of the event
+   */
+  public void addEvent(Date eventStart, Date eventEnd) {
+    this.morningFreeInterval = updateInterval(this.morningFreeInterval, eventStart, eventEnd);
+    this.workFreeInterval = updateInterval(this.workFreeInterval, eventStart, eventEnd);
+    this.eveningFreeInterval = updateInterval(this.eveningFreeInterval, eventStart, eventEnd);
+  }
+
+  /**
+   * Method to update the free intervals in a particular list of intervals.
```
antmar, line 77: Be more specific about what "update" means.

src/main/java/com/google/sps/data/CalendarDataResponse.java
```diff
@@ -0,0 +1,54 @@
...
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.util.*;
+
+/** Class containing the response to be converted to Json. */
+public final class CalendarDataResponse {
+
+  final long startDay;
```
antmar, line 22: What are the units of `startDay`?

src/main/java/com/google/sps/data/CalendarDataResponse.java
```diff
@@ -0,0 +1,54 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.util.*;
+
+/** Class containing the response to be converted to Json. */
+public final class CalendarDataResponse {
```
antmar, line 20: Call this something like `CalendarSummaryResponse`. "Date" is not very specific.

src/main/java/com/google/sps/data/CalendarDataResponse.java
```diff
@@ -0,0 +1,54 @@
...
+   */
+  public CalendarDataResponse(
+      long startDay, List<Long> workHoursPerDay, List<Long> personalHoursPerDay) {
+    this.startDay = startDay;
+    this.workHoursPerDay = workHoursPerDay;
+    this.personalHoursPerDay = personalHoursPerDay;
+  }
+
+  public long getStartDay() {
+    // Useful for testing purposes
```
antmar, line 41: I am not sure what you mean by "useful for testing purposes". These are just standard getters. You can remove these comments.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,142 @@
...
+   *
+   * @param freeTime the list of intervals to update.
+   * @param eventStart the start time of the new event
+   * @param eventEnd the end time of the new event
+   * @return the new list of intervals
+   */
+  private List<DatePair> updateInterval(List<DatePair> freeTime, Date eventStart, Date eventEnd) {
+    List<DatePair> updatedTime = new ArrayList<>();
+    for (DatePair interval : freeTime) {
+      if (interval.getKey().before(eventStart) && interval.getValue().after(eventEnd)) {
```
antmar, line 87: Add a brief comment documenting each case here. Also document what the unwritten "else" case is.

src/main/webapp/index.html
antmar: You are using the same classes for each day, so I don't think there would be a problem with generating this from Javascript.

src/main/webapp/populateData.js
```diff
@@ -162,6 +165,22 @@ function populateCalendar() {
...
 
 /**
+ * Convert the time in hours and minutes,
+ * and concatenate the string to be displayed
+ * @param {int} timeMilli the time in milliseconds that needs to be converted
+ * @return {String} String representing the converted free time
+ */
+function convertTime(timeMilli) {
+  const hours = Math.floor(timeMilli / (60 * 60 * 1000));
+  const minutes = Math.floor((timeMilli - hours * 60 * 60 * 1000)/(60*1000));
```
antmar, line 175: You could clean this up by defining some constants for the unit conversions.

For the minutes calculation, it would be cleaner to use `(timeMilli % HOUR_IN_MILLIS)` instead of subtracting hours.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
antmar: The overall algorithm is now much easier to understand - thank you.

src/main/webapp/populateData.js
```diff
@@ -162,6 +165,22 @@ function populateCalendar() {
 }
 
 /**
+ * Convert the time in hours and minutes,
+ * and concatenate the string to be displayed
+ * @param {int} timeMilli the time in milliseconds that needs to be converted
+ * @return {String} String representing the converted free time
+ */
+function convertTime(timeMilli) {
```
antmar, line 173: It may be simpler to have the servlet provide the data in separate hours and minutes. At least that way you could unit test this logic.

austinteshuba:
Enhancement: Whenever you can, can you also implement default values while the panel is loading (Reference changes in PR #140). Basically, at the start of the `populate-calendar` function, just set the values for each of the containers to `-` or something to signify no data is (yet) available. 

This isn't worth putting us off track for (so please feel free to let me know if you don't have time for this), but I think it would be cool

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -21,154 +21,122 @@
...
+  private static final long WORK_BEGIN = 10 * HOUR_MILLI;
+  private static final long WORK_END = 18 * HOUR_MILLI;
+  private static final long PERSONAL_END = 23 * HOUR_MILLI;
+  private static final int OFFSET = 7;
 
   /**
-   * Initialize the class by calculating the start day. The free hours should be eight hours for the
-   * working hours and for the pesonal time initially. Adjust the free hours on the first day to
-   * account for the fact that there are hours that have already passed
+   * Initialize the class with the start day. The work hours are harrrd-coded between 10 AM and 6
```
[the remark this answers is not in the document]
AymarMuhikira: Thank you, `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -21,154 +21,122 @@
...
-  private final long personalBegin = 7 * Hour;
-  private final long workBegin = 10 * Hour;
-  private final long workEnd = 18 * Hour;
-  private final long personalEnd = 23 * Hour;
+  private final Date startDate;
+  private List<Long> workHoursPerDay;
+  private List<Long> personalHoursPerDay;
+  private List<DatePair> morningFreeInterval;
+  private List<DatePair> workFreeInterval;
+  private List<DatePair> eveningFreeInterval;
```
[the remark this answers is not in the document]
AymarMuhikira: Changed these variables based on Anton's comments. The reason they were not declared as final is that the values can be changed when running addEvent. `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
AymarMuhikira: Used this. `Resolved`

src/main/java/com/google/sps/utility/DatePair.java
```diff
@@ -0,0 +1,43 @@
...
+
+package com.google.sps.utility;
+
+import java.util.*;
+
+/** Class implementing a Pair Data Structure. */
+public final class DatePair {
+
+  private final Date key;
+  private final Date value;
```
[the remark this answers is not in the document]
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/utility/DatePair.java
```diff
@@ -0,0 +1,43 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.utility;
+
+import java.util.*;
+
+/** Class implementing a Pair Data Structure. */
+public final class DatePair {
```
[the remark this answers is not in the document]
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,142 @@
...
+  private static final long HOUR_MILLI = TimeUnit.HOURS.toMillis(1);
+  private static final long DAY_MILLI = TimeUnit.DAYS.toMillis(1);
+  private static final long PERSONAL_BEGIN = 7 * HOUR_MILLI;
+  private static final long WORK_BEGIN = 10 * HOUR_MILLI;
+  private static final long WORK_END = 18 * HOUR_MILLI;
+  private static final long PERSONAL_END = 23 * HOUR_MILLI;
```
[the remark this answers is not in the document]
AymarMuhikira: Agreed, changed this. `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,142 @@
...
+  private static final int OFFSET = 7;
```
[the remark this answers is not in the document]
AymarMuhikira: Made the name more descriptive. It is related to the calculation of the index when the start day is later in the week than the interval day. `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * Initialize the class with the start day. The work hours are harrrd-coded between 10 AM and 6
+   * PM. The rest of the free hours are between 7 AM and 11 PM.
+   *
+   * @param startTime parameter that gives the time of the start/now.
+   */
+  public FreeTimeUtility(Date startDate) {
+    this.startDate = startDate;
+    Date basisDate = new Date(startDate.getYear(), startDate.getMonth(), startDate.getDate(), 0, 0);
```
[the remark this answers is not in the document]
AymarMuhikira: I looked into Calendar. I didn't find a way to use it to simplify the code, I would still have to get/give these parameters with the class.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -0,0 +1,142 @@
...
+
+    this.morningFreeInterval = new ArrayList<>();
+    this.workFreeInterval = new ArrayList<>();
+    this.eveningFreeInterval = new ArrayList<>();
+
+    for (int day = 0; day < 5; day++) {
```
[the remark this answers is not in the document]
AymarMuhikira: Made it a parameter. `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
AymarMuhikira: I agree. `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
AymarMuhikira: Changed the description. `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/data/CalendarDataResponse.java
AymarMuhikira: Changed to int. startDay is a value between 0 and 6 giving the day of the week on which to start displaying on the client side. `Resolved`

src/main/java/com/google/sps/data/CalendarDataResponse.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/data/CalendarDataResponse.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
AymarMuhikira: `Resolved`

src/main/webapp/populateData.js
AymarMuhikira: Moved this logic to the server side. I agree that the modulo would be better than the subtraction. We still need to divide by one minute in millisecods, otherwise the value would be minutes, seconds and milliseconds. `Resolved`

src/main/webapp/populateData.js
AymarMuhikira: `Resolved`

src/main/webapp/populateData.js
```diff
@@ -143,17 +143,36 @@ function populateCalendar() {
...
-              convertTime(hoursJson.workHoursPerDay[day]) + ' (working)';
-          const personalContainer =
-              document.querySelector('#personal-day-'+day);
-          personalContainer.innerText =
-              convertTime(hoursJson.personalHoursPerDay[day]) + ' (personal)';
+        const days = ['Su', 'Mo', 'Tu', 'We', 'Th', 'Fr', 'Sa'];
+        const panelContent = document.querySelector('#panel-content');
+        panelContent.innerHTML = '';
+        for (const day in hoursJson.workTimeFree) {
+          if (typeof day == 'string') {
```
antmar, line 150: What does it mean if the `day` is not a string?

src/main/webapp/populateData.js
```diff
@@ -143,17 +143,36 @@ function populateCalendar() {
...
+            personalContainer.className = 'u-block';
+            personalContainer.innerText =
+                hoursJson.personalTimeFree[day].hours +
+                'h ' + hoursJson.personalTimeFree[day].minutes +
+                'm free (personal)';
+            timeContainer.appendChild(workContainer);
+            timeContainer.appendChild(personalContainer);
+            panelContentEntry.appendChild(dayContainer);
+            panelContentEntry.appendChild(timeContainer);
+            panelContent.appendChild(panelContentEntry);
```
antmar, line 174: It may be simpler to just use a template literal[1] and write out the desired HTML in its source form, then append it to the `panelContent.innerHTML`. However, this approach is technically correct - it's just hard to understand.

You are welcome to submit as-is and file an issue to clean this up later.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

src/test/java/CalendarServletTest.java
```diff
@@ -40,32 +49,42 @@
...
-      String.format(
-          "[{},{\"summary\":\"%s\"},{\"summary\":\"%s\"}]", EVENT_SUMMARY_ONE, EVENT_SUMMARY_TWO);
+  private static final int OFFSET_YEAR = 1900;
+  // private static final String EVENT_ONE_TWO_JSON =
+  //     String.format(
+  //         "[{\"summary\":\"%s\"},{\"summary\":\"%s\"}]", EVENT_SUMMARY_ONE, EVENT_SUMMARY_TWO);
+  // private static final String EVENT_ALL_JSON =
+  //     String.format(
+  //         "[{},{\"summary\":\"%s\"},{\"summary\":\"%s\"}]", EVENT_SUMMARY_ONE,
+  // EVENT_SUMMARY_TWO);
```
antmar, line 59: Do you still need these constants? If not, please delete the comment.

src/test/java/CalendarServletTest.java
```diff
@@ -76,54 +95,75 @@ public void setUp() throws Exception {
...
+          (int) personalHoursPerDay.get(index), actual.getPersonalTimeFree().get(index).getHours());
+      Assert.assertEquals(0, actual.getPersonalTimeFree().get(index).getMinutes());
+    }
   }
 
   @Test
   public void twoCalendars() throws Exception {
     // Test case where there are two calendars with a defined event in each
-    // Events must be returned in order of retrieval
+    // Event with 1 working hour and event with 2 personal hours.
```
antmar, line 139: This comment is a little misleading because the 2nd event only overlaps with personal time for 1 hour. I think the test is actually correct, and the comment just needs to be updated.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -14,49 +14,77 @@
...
+
+  private List<DateInterval> workFreeInterval;
+  private List<DateInterval> eveningFreeInterval;
+  private static final int DAY_ADJUST = 7;
 
   /**
-   * Initialize the class with the start day. The work hours are harrrd-coded between 10 AM and 6
-   * PM. The rest of the free hours are between 7 AM and 11 PM.
+   * Initialize the class with the start day. The work hours are hard-coded between 10 AM and 6 PM.
+   * The rest of the free hours are between 7 AM and 11 PM.
```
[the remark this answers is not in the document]
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -14,49 +14,77 @@
...
+  /**
+   * The lists of time intervals from now (startDate) to the end day. They are divided by morning,
+   * work, and evening periods to allow easy computation of free hours, and later to create events
+   * during work/morning/evening hours
+   */
+  private List<DateInterval> morningFreeInterval;
+
+  private List<DateInterval> workFreeInterval;
+  private List<DateInterval> eveningFreeInterval;
+  private static final int DAY_ADJUST = 7;
```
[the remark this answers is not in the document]
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -14,49 +14,77 @@
...
+  private final int numDays;
+  /**
+   * The lists of time intervals from now (startDate) to the end day. They are divided by morning,
+   * work, and evening periods to allow easy computation of free hours, and later to create events
+   * during work/morning/evening hours
+   */
+  private List<DateInterval> morningFreeInterval;
+
+  private List<DateInterval> workFreeInterval;
+  private List<DateInterval> eveningFreeInterval;
```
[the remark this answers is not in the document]
AymarMuhikira: I added a comment describing what numDays is. I think saying that the length of the lists is numDays would be misleading as the size can change when adding events. For example, if an event occupies a whole day, the corresponding iintervals are removed. `Resolved`

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -14,49 +14,77 @@
...
-      Date personalEnd = new Date(basisDate.getTime() + day * DAY_MILLI + PERSONAL_END);
-      morningFreeInterval.add(new DatePair(personalStart, workStart));
-      workFreeInterval.add(new DatePair(workStart, workEnd));
-      eveningFreeInterval.add(new DatePair(workEnd, personalEnd));
+    for (int day = 0; day < numDays; day++) {
+      Date workStart =
+          Date.from(
+              basisDate
+                  .toInstant()
+                  .plus(Duration.ofDays(day))
```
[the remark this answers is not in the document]
AymarMuhikira: `Resolved`

src/main/webapp/populateData.js
AymarMuhikira: I used this check as day is given as a string from the response. I did not chose to make this check, but when using a `for-in` loop, the `lint.js` check requires to make a check for the type of the variable or else the check is not passed.

src/main/webapp/populateData.js
AymarMuhikira: I will file an issue to clean this up later.

src/test/java/CalendarServletTest.java
AymarMuhikira: `Resolved`

src/test/java/CalendarServletTest.java
AymarMuhikira: Thank you, `Resolved`

src/main/webapp/populateData.js
antmar: Weird, but ok.

src/main/java/com/google/sps/utility/FreeTimeUtility.java
gbuenoandrade: I agree it looks much cleaner. Thanks for working on this!

gbuenoandrade: RESOLVED

Issue #97, opened 2020-07-21
Repair authentication flow to avoid duplicate server requests


austinteshuba:
When a user is already signed in, the client requests information from the server twice. 

This harms performance.

PR #102, opened 2020-07-22
Create functionality for posting/getting tasklists

austinteshuba:
Add functionality to allow new tasklists to be posted, and for existing tasklists to be returned (alongside the tasks that the tasklist owns).

Functions created in client to make these requests. A method is currently added to the client that will 1) add a new tasklist (with the name = the current date (in current time zone) in UTC time and 2) get all of the tasklists and print them to console. 

Currently this prints out twice (and makes two new tasklists), due to issue #97. This will be resolved in a later PR

Console Output (See two empty tasklists): 
Screenshot 2020-07-22 at 3 45 40 PM

Google tasks view (see two tasklists with timestamps as names): 
Screenshot 2020-07-22 at 3 45 46 PM

Dependent on PR #106


austinteshuba:
@austinteshuba after MVP, base this off of master and handle the merge conflicts. 

Also, consider making a PR for a super class for the tasks tests to hold the constants.

src/main/java/com/google/sps/servlets/TaskListServlet.java
```diff
@@ -0,0 +1,136 @@
...
+   *
+   * @param taskLists list of TaskList objects
+   * @param tasksClient TasksClient implementation with valid Google credential
+   * @return HashMap where keys are taskList ids and the values are all the tasks those tasklists
+   *     contain
+   * @throws IOException if there is an issue with the TasksService
+   */
+  private HashMap<String, List<Task>> mapTaskListsToTasks(
+      List<TaskList> taskLists, TasksClient tasksClient) throws IOException {
+    HashMap<String, List<Task>> taskListsWithTasks = new HashMap<>();
```
tttzach, line 100: Good use of data structures!

src/main/webapp/populateData.js
```diff
@@ -15,8 +15,14 @@
...
 
 /* eslint-disable no-unused-vars */
-/* global signOut, AuthenticationError */
+/* global signOut, AuthenticationError, getDateInLocalTimeZone */
 // TODO: Refactor so populate functions are done in parallel (Issue #26)
+
+// Stores the last retrieved copy of the user's tasklists and tasks
+// (mapped by tasklistId)
+let tasklists = [];
+let tasks = {};
```
tttzach, line 24: Would it be better to have tasklists as {} and have each tasklist map to a [] of tasks?

src/main/java/com/google/sps/servlets/TaskListServlet.java
```diff
@@ -0,0 +1,136 @@
...
+    HashMap<String, List<Task>> taskListsWithTasks = mapTaskListsToTasks(taskLists, tasksClient);
+
+    Gson gson = new Gson();
+    JsonObject jsonObject = new JsonObject();
+    jsonObject.add("tasklists", gson.toJsonTree(taskLists));
+    jsonObject.add("tasks", gson.toJsonTree(taskListsWithTasks));
+    String json = gson.toJson(jsonObject);
+
+    response.setContentType("application/json");
+    response.getWriter().println(json);
```
tttzach, line 86: I think you can use JsonUtility here as well.

src/test/java/TaskListServletTest.java
```diff
@@ -0,0 +1,137 @@
...
+    Assert.assertEquals(400, response.getStatus());
+  }
+
+  @Test
+  public void postTasklist() throws Exception {
+    Mockito.when(request.getParameter("taskListTitle")).thenReturn(VALID_TASKLIST_TITLE);
+    Mockito.when(tasksClient.postTaskList(VALID_TASKLIST_TITLE)).thenReturn(validTaskList);
+
+    servlet.doPost(request, response);
+    Assert.assertTrue(stringWriter.toString().contains(VALID_TASKLIST_JSON));
```
tttzach, line 135: I would recommend decoding the JSON and directly comparing it with `validTaskList`. Same for the other tests.

src/main/java/com/google/sps/servlets/TaskListServlet.java
austinteshuba: Thank you!

src/main/webapp/populateData.js
austinteshuba: Really interesting question actually - so this is possible in Java with hashmaps, but its impossible in JavaScript with objects. 

Objects need to have a key that's a string - not an entire object. I could build a custom data structure that lets you get the tasks from a specific tasklist, but under the hood it would essentially boil down to this. 

I'll probably take a look at making this cleaner later (maybe even building that data structure), but for now I just wanted to demonstrate that the server side changes work by implementing a simple implementation on the client. 

Open for suggestions though if you have an easy way to associate objects with a list of objects in JS

src/main/java/com/google/sps/servlets/TaskListServlet.java
austinteshuba: Good idea - I got thrown off by using JsonElement. Thanks!

src/test/java/TaskListServletTest.java
austinteshuba: Agreed - fixing now! Thanks!

src/main/java/com/google/sps/servlets/TaskListServlet.java
austinteshuba: I unfortunately cannot - gson needs to know that the object passed is of type `JsonObject` as it applies different serialization methods to that type. I logged an issue to extend the JsonUtility at a later date (#107)

src/test/java/TaskListServletTest.java
austinteshuba: Update - I agree this would be ideal, but gson is giving me a ton of trouble when I try to do this. `TaskList` does not play well with gson when trying to deserialize JSON (Even when I do `TypeToken<TaskList>(){}.getType()`, I still get an error saying `java.lang.IllegalArgumentException: Map objects need to be parameterized unless you use a custom serializer. Use the com.google.gson.reflect.TypeToken to extract the ParameterizedType.`). 

I'm not sure how to proceed, and I want to check in with the team before working on this too much.

tttzach: Yes please do, there's other things you need to pull your hair over and I don't think this is one of them.

src/main/java/com/google/sps/servlets/TaskListServlet.java
tttzach: Alright, fair enough! Thanks!

src/main/webapp/populateData.js
tttzach: Ok cool! I agree that it's probably going to be [less clean](https://stackoverflow.com/questions/36299664/javascript-convert-a-list-of-objects-to-an-associative-array) if you were to associate the objects with a list of objects in JS.

src/main/webapp/populateData.js
```diff
@@ -154,3 +160,76 @@ function populateCalendar() {
...
+              console.log(taskLists);
+              console.log(tasks);
+            });
+      });
+}
+
+/**
+ * Update the tasks and taskLists lists.
+ *
+ * @return {Promise<any>} A promise that is resolved once the tasks and
```
gbuenoandrade, line 186: Where? (same below)

src/test/java/TaskListServletTest.java
```diff
@@ -0,0 +1,138 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.Mockito;
+
+/**
+ * Test that the Tasks Servlet responds to the client with correctly parsed Task names. Assumes
+ * AuthenticatedHttpServlet is functioning properly (those tests will fail otherwise).
```
gbuenoandrade, line 34: Isn't this the default assumption for any test?

src/main/webapp/populateData.js
austinteshuba: `fetch(url)` is a promise - I returned the entire fetch block, which is a promise

src/test/java/TaskListServletTest.java
austinteshuba: Fair enough - removing

src/main/java/com/google/sps/servlets/TaskListServlet.java
```diff
@@ -0,0 +1,138 @@
...
+  /**
+   * Map taskLists to the tasks they contain
+   *
+   * @param taskLists list of TaskList objects
+   * @param tasksClient TasksClient implementation with valid Google credential
+   * @return HashMap where keys are taskList ids and the values are all the tasks those taskLists
+   *     contain
+   * @throws IOException if there is an issue with the TasksService
+   */
+  private HashMap<String, List<Task>> mapTaskListsToTasks(
```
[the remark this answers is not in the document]
austinteshuba: Oops - fixed (also fixed in other areas of the code where applicable)

src/main/webapp/populateData.js
gbuenoandrade: ACK

PR #103, opened 2020-07-22
Allow for posting new tasks to user's account

austinteshuba:
Includes client side function to post a new task, as well as a sample implementation where a sample task is added to a new task list on login. 

Note that this occurs twice due to Issue #97.

Console response: 
Screenshot 2020-07-22 at 8 20 55 PM

Sample task created:
Screenshot 2020-07-22 at 8 21 23 PM

Dependent on PR #102


src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -61,13 +62,76 @@ public TasksServlet(
...
+    TasksResponse tasksResponse =
+        new TasksResponse(
+            taskListTitles, tasksToComplete, tasksDueToday, tasksCompletedToday, tasksOverdue);
+
+    // Convert tasks to JSON and print to response
+    Gson gson = new Gson();
+    String tasksJson = gson.toJson(tasksResponse);
+
+    response.setContentType("application/json");
+    response.getWriter().println(tasksJson);
```
tttzach, line 100: This is a good candidate for `JsonUtility`.

src/test/java/TasksServletTest.java
```diff
@@ -205,12 +209,42 @@ public void allTasks() throws Exception {
...
+    BufferedReader bufferedReader = new BufferedReader(reader);
+    Mockito.when(request.getReader()).thenReturn(bufferedReader);
+
+    servlet.doPost(request, response);
+
+    Task postedTask = gson.fromJson(stringWriter.toString(), Task.class);
+
+    Assert.assertEquals(validTask.getTitle(), postedTask.getTitle());
+    Assert.assertEquals(validTask.getNotes(), postedTask.getNotes());
+    Assert.assertEquals(validTask.getDue(), postedTask.getDue());
```
tttzach, line 248: Propagating Anton's feedback on my TasksResponse tests - you can override equals in Task to compare these two directly.

src/main/java/com/google/sps/servlets/TasksServlet.java
austinteshuba: Agreed - done!

src/test/java/TasksServletTest.java
austinteshuba: Task is a final, external class - I'm not sure how I would override this, or if it's a good idea to try (generally overriding external dependencies is bad practice AFAIK)

tttzach: Right - my bad. I was assuming Task was an internal class. Thanks for pointing this out!

src/test/java/TasksServletTest.java
```diff
@@ -205,12 +208,42 @@ public void allTasks() throws Exception {
     Mockito.when(tasksClient.listTaskLists()).thenReturn(ONE_TASK_LIST);
     Mockito.when(tasksClient.listTasks(TASK_LIST_ONE)).thenReturn(ALL_TASKS);
     servlet.doGet(request, response);
-
```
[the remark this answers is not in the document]
austinteshuba: I believe I made this more consistent - please let me know if there's still an issue!

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -61,13 +62,72 @@ public TasksServlet(
...
+      return;
+    }
+
+    Gson gson = new Gson();
+    Task taskToPost = gson.fromJson(request.getReader(), Task.class);
+
+    try {
+      JsonUtility.sendJson(response, tasksClient.postTask(taskListId, taskToPost));
+    } catch (IOException e) {
+      response.sendError(400, "taskListId is invalid");
```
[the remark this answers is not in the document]
austinteshuba: Yeah good catch - so an invalid `taskListId` can cause an IOException here (Google does not have this documented - I tested this manually), however there are also tons of other reasons this could happen. 

I changed this to throw the error with a different message for the client, that indicates the `taskListId` could be the cause, but not the only cause. Let me know what you think!

src/test/java/TasksServletTest.java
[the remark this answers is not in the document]
gbuenoandrade: Looks good, tks!
RESOLVED

src/main/java/com/google/sps/servlets/TasksServlet.java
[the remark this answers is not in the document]
gbuenoandrade: It's good, thanks for improving this.
RESOLVED

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -61,13 +62,73 @@ public TasksServlet(
...
+
+    String taskListId = request.getParameter("taskListId");
+
+    if (taskListId == null || taskListId.equals("")) {
+      response.sendError(400, "taskListId must be present in request");
+      return;
+    }
+
+    Gson gson = new Gson();
+    Task taskToPost = gson.fromJson(request.getReader(), Task.class);
```
antmar, line 124: What happens if the body is empty? Or if the task can't be parsed?

(Maybe you don't need any special handling, but at least add a test to show what the behavior is.)

austinteshuba: I added some handling for this. Thanks for catching this!

PR #105, opened 2020-07-23
Refactor Tasks panel

tttzach:
This PR addresses the outstanding review comments in https://github.com/googleinterns/step7-2020/pull/88 which was merged prematurely for our MVP demo.


src/main/java/com/google/sps/model/TasksResponse.java
```diff
@@ -14,61 +14,21 @@
...
 
+import com.google.appengine.repackaged.com.google.common.collect.ImmutableList;
 import java.util.List;
+import lombok.Builder;
 
 /**
  * Contains the summary tasks information that should be passed to the client, as well as the
  * methods to generate these statistics.
  */
+@Builder
```
tttzach, line 25: https://github.com/googleinterns/step7-2020/pull/88#discussion_r459092523
> This constructor is just at the edge of being too complicated. If you were to add any more fields, I would recommend switching to a Builder pattern for initializing this object.
The concern here is that it would be very easy to get the order of the parameters wrong, and there is no way to see that at the call site. Unit tests should help catch bugs, but it's better when the code is self-documenting.

src/test/java/TasksServletTest.java
```diff
@@ -128,124 +205,98 @@ public void setUp() throws IOException, GeneralSecurityException {
...
     servlet.doGet(request, response);
-    printWriter.flush();
-    TasksResponse tasksResponse = gson.fromJson(stringWriter.toString(), TasksResponse.class);
-    Assert.assertEquals(NO_TASKLISTS_TITLES, tasksResponse.getTaskListNames());
-    Assert.assertEquals(0, tasksResponse.getTasksToComplete());
-    Assert.assertEquals(0, tasksResponse.getTasksDueToday());
-    Assert.assertEquals(0, tasksResponse.getTasksCompletedToday());
-    Assert.assertEquals(0, tasksResponse.getTasksOverdue());
+    TasksResponse actual = gson.fromJson(stringWriter.toString(), TasksResponse.class);
+    Assert.assertEquals(NO_TASKLISTS_RESPONSE, actual);
```
tttzach, line 227: https://github.com/googleinterns/step7-2020/pull/88#discussion_r459098546
> You could override `Object.equals` in `TasksResponse`, and then use `Assert.assertEquals` to compare `TasksResponse` objects directly.

src/test/java/TasksServletTest.java
```diff
@@ -116,8 +120,81 @@
...
+          .tasksOverdue(1)
+          .build();
+  private static final TasksResponse MULTIPLE_TASKLISTS_RESPONSE =
+      TasksResponse.builder()
+          .taskListTitles(TWO_TASKLISTS_TITLES)
+          .tasksToComplete(6)
+          .tasksDueToday(2)
+          .tasksCompletedToday(2)
+          .tasksOverdue(2)
+          .build();
```
tttzach, line 194: https://github.com/googleinterns/step7-2020/pull/88#discussion_r459100105
> It's not immediately obvious that these responses are correct, just by looking at this test method. I see that you have incomplete and overdue tasks, but how do I know that `TASKS_DUE_YESTERDAY` only returns one overdue task?
> 
> If you follow my suggestion above (to allow comparing responses), then you can define static constants for the expected responses to each set of input tasks. This would put the definition of the responses closer to where the inputs are defined.
> 
> (I am not sure what you would do with more complex tests that use multiple task lists - probably just define responses for those combinations.)

src/test/java/TasksServletTest.java
```diff
@@ -128,124 +205,98 @@ public void setUp() throws IOException, GeneralSecurityException {
...
-    Assert.assertEquals(3, tasksResponse.getTasksToComplete());
-    Assert.assertEquals(1, tasksResponse.getTasksDueToday());
-    Assert.assertEquals(1, tasksResponse.getTasksCompletedToday());
-    Assert.assertEquals(1, tasksResponse.getTasksOverdue());
+    TasksResponse actual = gson.fromJson(stringWriter.toString(), TasksResponse.class);
+    Assert.assertEquals(ALL_TASKS_RESPONSE, actual);
+  }
+
+  @Test
+  public void multipleTaskLists() throws Exception {
```
tttzach, line 294: https://github.com/googleinterns/step7-2020/pull/88#discussion_r459100885
> You should add a test with more than one task list.

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -74,56 +74,48 @@ public TasksServlet(
...
+    return tasks.stream()
+        .filter(task -> task.getDue() != null)
+        .filter(
+            task -> {
+              DateTime dateTime = DateTime.parseRfc3339(task.getDue().replace("Z", zoneOffset));
+              long millis = dateTime.getValue();
+              Instant dueDate = new Date(millis).toInstant().plus(Period.ofDays(1));
+              return dueDate.isBefore(endOfDay);
+            })
+        .count();
```
tttzach, line 118: https://github.com/googleinterns/step7-2020/pull/88#discussion_r459058956
> Maybe you could still use .count()?

src/test/java/TasksServletTest.java
```diff
@@ -128,124 +205,98 @@ public void setUp() throws IOException, GeneralSecurityException {
...
-
     // Writer used in get/post requests to capture HTTP response values
     stringWriter = new StringWriter();
     printWriter = new PrintWriter(stringWriter);
+
+    request = Mockito.mock(HttpServletRequest.class);
+    response =
+        Mockito.mock(
+            HttpServletResponse.class,
+            AdditionalAnswers.delegatesTo(new HttpServletResponseFake(stringWriter)));
```
tttzach, line 216: https://github.com/googleinterns/step7-2020/pull/88#discussion_r459060737
> Please use Austin's fake response when #86 lands

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: I have refactored all the explicit for loops into streams to maintain consistency.

antmar: Nice!

src/test/java/TasksServletTest.java
```diff
@@ -128,124 +205,98 @@ public void setUp() throws IOException, GeneralSecurityException {
...
+
+    Mockito.when(request.getCookies()).thenReturn(validCookies);
     Mockito.when(response.getWriter()).thenReturn(printWriter);
```
antmar, line 219: I don't think you need to mock `response.getWriter()` any more.

src/test/java/TasksServletTest.java
antmar: Much more readable now!

src/test/java/TasksServletTest.java
tttzach: Yup you're right - it's in `HttpServletResponseFake` already, thanks for catching that!

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: Thanks!

src/test/java/TasksServletTest.java
tttzach: Thanks!

src/main/java/com/google/sps/model/TasksResponse.java
tttzach: `RESOLVED`

src/test/java/TasksServletTest.java
tttzach: `RESOLVED`

src/test/java/TasksServletTest.java
tttzach: `RESOLVED`

src/test/java/TasksServletTest.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: `RESOLVED`

src/test/java/TasksServletTest.java
tttzach: `RESOLVED`

src/test/java/TasksServletTest.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -146,56 +151,48 @@ public void doPost(
...
-    }
-    return tasksCompletedToday;
+    // getHidden is defined for completed tasks
+    return tasks.stream()
+        .filter(task -> task.getHidden() != null)
+        .filter(
+            task -> {
+              Instant completionDateTime = ZonedDateTime.parse(task.getUpdated()).toInstant();
+              return completionDateTime.isAfter(startOfDay)
+                  && completionDateTime.isBefore(endOfDay);
```
gbuenoandrade, line 177: This looks pretty cool!

tttzach: Thanks!

PR #106, opened 2020-07-23
Create superclasses for testing servlets

austinteshuba:
Create AuthenticatedServletTestBase and ServletTestBase to make creating servlet tests easier. 

Also switched throws to generic "Exception" class where possible for consistency.

### DIFF
--- a/src/test/java/CalendarServletTest.java
+++ b/src/test/java/CalendarServletTest.java
@@ -17,20 +17,11 @@ import com.google.api.services.calendar.model.Event;
 import com.google.appengine.repackaged.com.google.gson.Gson;
 import com.google.common.collect.ImmutableList;
 import com.google.gson.reflect.TypeToken;
-import com.google.sps.model.AuthenticationVerifier;
 import com.google.sps.model.CalendarClient;
 import com.google.sps.model.CalendarClientFactory;
 import com.google.sps.servlets.CalendarServlet;
-import java.io.IOException;
-import java.io.PrintWriter;
-import java.io.StringWriter;
 import java.lang.reflect.Type;
-import java.security.GeneralSecurityException;
 import java.util.List;
-import javax.servlet.ServletException;
-import javax.servlet.http.Cookie;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
@@ -40,26 +31,12 @@ import org.mockito.Mockito;
 
 /** Test Calendar Servlet responds to client with correctly parsed Events. */
 @RunWith(JUnit4.class)
-public final class CalendarServletTest {
-  private AuthenticationVerifier authenticationVerifier;
+public final class CalendarServletTest extends AuthenticatedServletTestBase {
   private CalendarClientFactory calendarClientFactory;
   private CalendarClient calendarClient;
   private CalendarServlet servlet;
-  private HttpServletRequest request;
-  private HttpServletResponse response;
-  private StringWriter stringWriter;
-  private PrintWriter printWriter;
-  private static final Gson gson = new Gson();
 
-  private static final String ID_TOKEN_KEY = "idToken";
-  private static final String ID_TOKEN_VALUE = "sampleId";
-  private static final String ACCESS_TOKEN_KEY = "accessToken";
-  private static final String ACCESS_TOKEN_VALUE = "sampleAccessToken";
-  private static final Cookie sampleIdTokenCookie = new Cookie(ID_TOKEN_KEY, ID_TOKEN_VALUE);
-  private static final Cookie sampleAccessTokenCookie =
-      new Cookie(ACCESS_TOKEN_KEY, ACCESS_TOKEN_VALUE);
-  private static final Cookie[] validCookies =
-      new Cookie[] {sampleIdTokenCookie, sampleAccessTokenCookie};
+  private static final Gson gson = new Gson();
 
   private static final String EVENT_SUMMARY_ONE = "test event one";
   private static final String EVENT_SUMMARY_TWO = "test event two";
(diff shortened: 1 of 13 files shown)


src/test/java/CalendarServletTest.java
```diff
@@ -90,29 +67,19 @@
...
-
-    // Writer used in get/post requests to capture HTTP response values
-    stringWriter = new StringWriter();
-    printWriter = new PrintWriter(stringWriter);
-    Mockito.when(response.getWriter()).thenReturn(printWriter);
   }
 
   @Test
-  public void noCalendarEvent() throws IOException, ServletException {
+  public void noCalendarEvent() throws Exception {
```
tttzach, line 82: I received this review comment from either Gui or Anton on one of my PRs - throwing IOException and ServletException like before is better because it's more descriptive than just Exception.

src/test/java/ClientIDServletTest.java
```diff
@@ -14,42 +14,19 @@
...
-    response = Mockito.mock(HttpServletResponse.class);
-
-    stringWriter = new StringWriter();
-    printWriter = new PrintWriter(stringWriter);
-    Mockito.when(response.getWriter()).thenReturn(printWriter);
-  }
-
   @Test
-  public void responseContainsClientId() throws IOException {
+  public void responseContainsClientId() throws Exception {
```
tttzach, line 28: Same comment here about the exceptions.

src/test/java/JsonUtilityTest.java
```diff
@@ -60,7 +59,7 @@ public String revealSecret() {
   private final Gson gson = new Gson();
 
   @Before
-  public void setUp() throws IOException {
+  public void setUp() throws Exception {
```
tttzach, line 62: Same comment here about the exceptions.

src/test/java/KeyProviderTest.java
```diff
@@ -30,7 +29,7 @@
   private static final File file = new File("src/main/resources/TEST_KEYS.json");
 
   @BeforeClass
-  public static void setUp() throws IOException {
+  public static void setUp() throws Exception {
```
tttzach, line 32: Same comment here about the exceptions.

src/test/java/ServletTestBase.java
```diff
@@ -0,0 +1,40 @@
...
+ * Base test class for testing servlets. Does not contain authentication handling (use
+ * AuthenticatedServletTestBase for servlets that require user authentication
+ */
+public abstract class ServletTestBase {
+  protected HttpServletRequest request;
+  protected HttpServletResponse response;
+  protected StringWriter stringWriter;
+
+  @Before
+  public void setUp() throws Exception {
```
tttzach, line 32: Same comment here about the exceptions.

src/test/java/TasksServletTest.java
```diff
@@ -116,33 +91,23 @@
           TASK_COMPLETED_YESTERDAY,
           TASK_COMPLETED_TODAY);
 
+  @Override
   @Before
-  public void setUp() throws IOException, GeneralSecurityException {
-    authenticationVerifier = Mockito.mock(AuthenticationVerifier.class);
+  public void setUp() throws Exception {
```
tttzach, line 96: Same comment here about the exceptions.

src/test/java/TasksServletTest.java
```diff
@@ -116,33 +91,23 @@
...
-    Mockito.when(response.getWriter()).thenReturn(printWriter);
   }
 
   @Test
-  public void noTaskLists() throws IOException, ServletException {
+  public void noTaskLists() throws Exception {
     Mockito.when(tasksClient.listTaskLists()).thenReturn(NO_TASKLISTS);
     servlet.doGet(request, response);
-    printWriter.flush();
+
```
tttzach, line 110: Just curious because I might change this for my code as well - is `printWriter.flush()` not needed?

austinteshuba: Thanks for pointing this out - You can actually create a printWriter with autoFlush (which will flush the stream every time `printWriter.println()` is called. I just put this into the `HttpServletResponseFake` as I initially intended to (it wasn't there when you reviewed).

src/test/java/CalendarServletTest.java
austinteshuba: @gbuenoandrade informed me in a [PR](https://github.com/googleinterns/step7-2020/pull/86#discussion_r458278603) that you can just throw Exception in tests. 

This part of the refactor is far from needed, so I can undo this if anyone is adamantly against it, but if it's permissible for tests then I don't see a reason not to.

src/test/java/TasksServletTest.java
tttzach: Oh that's neat, thanks!

src/test/java/CalendarServletTest.java
tttzach: That's fair! I just noticed that the files where you changed these are all tests.

src/test/java/GmailServletTest.java
```diff
@@ -39,54 +35,27 @@
  * will fail otherwise).
  */
 @RunWith(JUnit4.class)
-public final class GmailServletTest extends GmailTestBase {
+public final class GmailServletTest extends AuthenticatedServletTestBase implements GmailTestBase {
```
gbuenoandrade, line 38: Apparently, implementing an interface that contains several constants is discouraged - [constant interface anti-pattern](https://stackoverflow.com/a/2659740/13744548).

src/test/java/GmailClientTest.java
```diff
@@ -21,7 +21,6 @@
 /** Test GmailClient static functions */
 @RunWith(JUnit4.class)
 public final class GmailClientTest {
-
```
gbuenoandrade, line 24: Nit: you can leave this as is, but for future reference, try to leave this kind of refactoring (same about `JsonUtilityTest.java`) for a PR that contains no logic. (Especially because both this and `JsonUtilityTest.java` don't contain any change in logic.)

austinteshuba: Got it - sorry about that!

src/test/java/GmailServletTest.java
austinteshuba: I agree - removing. Thanks!

austinteshuba:
> Although I'm not particularly concerned with this, it is worth pointing out that using inheritance in tests to reduce code repetition is a controversial topic. (See [this](https://www.petrikainulainen.net/programming/unit-testing/3-reasons-why-we-should-not-use-inheritance-in-our-tests/) for some reasons why you should not use it.) Composition is an alternative. I leave it to you to decide.
> 
> LGTM

Thanks for this - it was a good read. 

I think I'm going to keep the inheritance for a few reasons. If anyone has strong opinions later, it's simple enough to remove. 
I would argue that there is an "is a" relationship between `ServletTestBase` and `AuthenticatedServletTestBase` with their children classes. `ClientIDServletTest` is a type of servlet test, and `GmailServletTest` is a type of Authenticated servlet test, just how `ClientIDServlet` is an `HttpServlet` and `GmailServlet` is an `AuthenticatedHttpServlet`. 

In terms of making the tests slightly slower, I doubt we will have thousands of servlet tests (which is the scale required for this lag to matter), regardless of how many servlets we make. After all, the goal of testing servlets is to ensure they properly parse and generate responses (testing the servlet doesn't really concern itself with arriving at the right answer, just combining the responses from other algorithms correctly) - there are naturally fewer tests needed to do this each time.

Again this is pretty minor, and I don't care too much either way, but I'm personally not convinced that this is truly a misuse of inheritance. Please let me know later if you feel strongly the other way - I'll happily undo this.

PR #109, opened 2020-07-24
Update GmailClient to build search queries for words in a subject line

austinteshuba:
This eventually will be used for the Assign panel, where certain words will be queried in an email's subject line. 

I also eliminated the `emailQueryString()` method, as it was unneeded and confusing. It was only used once in the entire codebase to combine two query strings (not 6). I instead replaced it with a method that will concatenate any number of search queries (which are generated by the other methods in `GmailClient`) together.


austinteshuba:
Rebasing to be off master on a new branch.

PR #110, opened 2020-07-24
Update GmailClient query string methods

austinteshuba:
This eventually will be used for the Assign panel, where certain words will be queried in an email's subject line. 

I also eliminated the `emailQueryString()` method, as it was unneeded and confusing. It was only used once in the entire codebase to combine two query strings (not 6). I instead replaced it with a method that will concatenate any number of search queries (which are generated by the other methods in `GmailClient`) together.

Dependent on PR #113


src/test/java/GmailClientTest.java
```diff
@@ -131,15 +139,45 @@ public void getQueryStringDontFilterForImportantTag() {
...
-    String multipleFilterQuery =
-        GmailClient.emailQueryString(ONE_UNIT_OF_TIME, DAYS_UNIT, true, true, SAMPLE_EMAIL);
-
-    Assert.assertTrue(multipleFilterQuery.contains(ONE_DAY_QUERY));
-    Assert.assertTrue(multipleFilterQuery.contains(UNREAD_EMAILS_QUERY));
-    Assert.assertTrue(multipleFilterQuery.contains(IMPORTANT_EMAILS_QUERY));
-    Assert.assertTrue(multipleFilterQuery.contains(FROM_EMAIL_QUERY));
+  public void getQueryStringNoSubjectWords() {
+    String subjectLineQuery = GmailClient.wordsInSubjectLineQuery(EMPTY_SUBJECT_LINE_WORDS);
+
```
tttzach, line 144: very minor nit: this two-line test has a newline in between and the two-line test `combineQueriesNonePresent` doesn't, would suggest either removing this newline or adding a newline after line 168

austinteshuba: Consider it done!

austinteshuba:
Currently, the search query 'subject:("PHRASE ONE" OR "PHRASE TWO")' **appears** to do be the correct syntax to return emails with PHRASE ONE or PHRASE TWO in the subject line. [Source](https://support.google.com/mail/thread/14162941?hl=en). If you have a second, try this out in Gmail and let me know if this doesn't work for you - would like to validate this as much as possible.

src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -70,6 +71,17 @@
     }
   }
 
+  /**
+   * Given some search queries, combine them into a larger query Will not remove
```
antmar, line 75: Nit: I think you are missing a period before "Will".

src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -160,9 +144,28 @@ static String fromEmailQuery(String from) {
...
     return isImportant ? "is:important" : "";
   }
+
+  /**
+   * Creates Gmail query to find emails with the passed word in the subject line
+   *
+   * @param words words to check subject line for.
+   * @return query string to use in gmail to find emails that contain the words
+   */
+  static String wordsInSubjectLineQuery(List<String> words) {
```
antmar, line 158: Make it clear this query will match any of the words. I suggest changing the method name and updating the documentation. 

(Also, using the term "words" is a little too specific. You don't do anything with spaces or other word delimiters. Perhaps a list of "phrases" or "tokens" would be more general.)

src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -70,6 +71,17 @@
...
   }
 
+  /**
+   * Given some search queries, combine them into a larger query Will not remove
+   * duplicates/contradictions. TODO: Refactor query creation using the Builder pattern (Issue #100)
+   *
+   * @param queries gmail search queries
+   * @return String with queries space separated for use in Gmail
+   */
+  static String combineSearchQueries(String... queries) {
```
antmar, line 81: Make it clear that you are effectively combining the queries using AND operators.

src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -160,9 +144,28 @@ static String fromEmailQuery(String from) {
...
+   *
+   * @param words words to check subject line for.
+   * @return query string to use in gmail to find emails that contain the words
+   */
+  static String wordsInSubjectLineQuery(List<String> words) {
+    if (words.isEmpty()) {
+      return "";
+    }
+
+    // Surround each word with double quotes (""), if they are not already there
```
antmar, line 163: Put this information in the method Javadoc.

src/test/java/GmailClientTest.java
```diff
@@ -130,15 +140,54 @@ public void getQueryStringDontFilterForImportantTag() {
...
+                .substring(subjectLineQuery.indexOf('(') + 1, subjectLineQuery.indexOf(')'))
+                .split(" OR "));
+
+    // Expected words should be the same as the input, with quotes surrounding each word (if not
+    // already present)
+    // Ex: "WordOne", wordTwo -> "WordOne", "wordTwo"
+    List<String> expectedSubjectLineWords =
+        SOME_SUBJECT_LINE_WORDS.stream()
+            .map((word) -> String.format("\"%s\"", StringUtils.strip(word, "\"")))
+            .collect(Collectors.toList());
```
antmar, line 169: You don't need to recreate this logic in the test. You are providing the input data, so you know if the individual tokens are quoted or not. You should just have a separate test case to check that quotes are handled correctly.

In general, we want to make tests as simple as possible. We don't write tests for tests, so it's hard to verify correctness if the test logic is not trivial.

src/test/java/GmailClientTest.java
```diff
@@ -130,15 +140,54 @@ public void getQueryStringDontFilterForImportantTag() {
...
+
+    // Query should be in the form of "subject:("wordOne" OR "wordTwo" OR "wordThree")"
+    // The order of the words in the brackets does not matter.
+    String actualSubjectLineQueryPrefix =
+        subjectLineQuery.substring(0, subjectLineQuery.indexOf(':') + 1);
+    List<String> actualSubjectLineWords =
+        Arrays.asList(
+            subjectLineQuery
+                .substring(subjectLineQuery.indexOf('(') + 1, subjectLineQuery.indexOf(')'))
+                .split(" OR "));
```
antmar, line 161: You could use a regular expression (regex) to match the prefix and extract the words, all in a single method call.

Alternatively, if you want to assume the order is fixed, you could just match the exact string, or match a regex directly. (I don't think you are actually handling different word orders in this test, unless I am missing something?)

src/test/java/GmailClientTest.java
austinteshuba: Agreed - I'm just going to make the assumption here that the order of the words should be the same in the string (while they don't have to be, I don't think this is a bad assumption - there would be no reason to change the order) so I can compare the strings exactly.

src/test/java/GmailClientTest.java
austinteshuba: Agreed - I'm just going to make the assumption here that the order of the words should be the same in the string (while they don't have to be, I don't think this is a bad assumption - there would be no reason to change the order) so I can compare the strings exactly.

src/main/java/com/google/sps/model/GmailClient.java
austinteshuba: Done! Thanks!

src/main/java/com/google/sps/model/GmailClient.java
austinteshuba: Yep agreed. Thanks!

src/main/java/com/google/sps/model/GmailClient.java
austinteshuba: Agreed - fixed!

src/main/java/com/google/sps/model/GmailClient.java
austinteshuba: Oops - It was originally a new line, but spotless removed that. All fixed!

PR #112, opened 2020-07-24
Create servlet for assign panel

austinteshuba:
Prioritization algorithm still to come - for now just returns emails based on filtered query string.

Client side implementation also still to come. Will not merge this to master until client side is ready

Dependent on PR #110


src/test/java/GmailActionableEmailsServletTest.java
```diff
@@ -0,0 +1,179 @@
...
+    Mockito.when(gmailClientFactory.getGmailClient(Mockito.any())).thenReturn(gmailClient);
+  }
+
+  @Test
+  public void subjectLineWordsNull() throws Exception {
+    Mockito.when(request.getParameter("unreadOnly")).thenReturn(String.valueOf(true));
+    Mockito.when(request.getParameter("nDays")).thenReturn(String.valueOf(DEFAULT_N_DAYS));
+
+    servlet.doGet(request, response);
+    Assert.assertEquals(400, response.getStatus());
```
tttzach, line 87: If I understand correctly, 400 is to show that an error has occurred. I think it would be nice to have a comment explaining that since it's not immediately clear why an error is occurring.

src/test/java/GmailActionableEmailsServletTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+/** Tests the GmailActionableEmailsServlet */
+@RunWith(JUnit4.class)
+public class GmailActionableEmailsServletTest extends AuthenticatedServletTestBase {
+  private GmailClient gmailClient;
+  private GmailActionableEmailsServlet servlet;
+
+  private static final Gson gson = new Gson();
+
+  // spaces should be parsed correctly by request.getParameter if query string is properly encoded
```
tttzach, line 43: very minor nit: Spaces instead of spaces

src/main/java/com/google/sps/servlets/GmailActionableEmailsServlet.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Servlet to get actionable emails from user's gmail account, as defined by specific words in an
+ * email's subject line. Used by "Assign" panel on client.
+ */
+@WebServlet("/gmail-actionable-emails")
```
tttzach, line 37: I am not well versed on the design you decided on for the Assign panel - would it be easier if you had a single Assign servlet and have GmailActionableEmails as a client?

src/test/java/GmailActionableEmailsServletTest.java
austinteshuba: 400 is to show an illegal argument exception - the name of the method specifies that the subjectWordPhrases is null (which should imply the error here)

src/test/java/GmailActionableEmailsServletTest.java
austinteshuba: Fixed! Thanks!

src/main/java/com/google/sps/servlets/GmailActionableEmailsServlet.java
austinteshuba: I think maybe this is a naming issue - this is the only servlet used for getting the actionable emails from gmail (and the only servlet used for getting information from the panel). 

The other servlet used will be tasks, since I will be posting tasks when appropriate. I was thinking of grouping them together, but I don't think it makes sense semantically (posting tasks and getting actionable emails is two different functionalities - the server design should not be dictated by the client's use of the servlets). 

Just my two cents - let me know if this cleared it up, or if you think we should switch some names up on the servlets for clarity.

src/test/java/GmailActionableEmailsServletTest.java
tttzach: Fair enough - thanks for clarifying!

src/main/java/com/google/sps/servlets/GmailActionableEmailsServlet.java
tttzach: Yeah, I agree that it's probably the names that are a little confusing (same issue with my servlets and clients). I think it's fine to leave it for now, we can always refactor later when it's much more obvious if something needs to be renamed.

src/test/java/GmailActionableEmailsServletTest.java
tttzach: `RESOLVED`

src/test/java/GmailActionableEmailsServletTest.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GmailActionableEmailsServlet.java
austinteshuba: RESOLVED

src/main/java/com/google/sps/servlets/GmailActionableEmailsServlet.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+    if (nDays <= 0) {
+      response.sendError(400, "nDays must be positive");
+      return;
+    }
+
+    List<Message> actionableEmails =
+        gmailClient.getActionableEmails(
+            GmailClient.MessageFormat.FULL, subjectLinePhrases, unreadOnly, nDays);
+    JsonUtility.sendJson(response, actionableEmails);
```
antmar, line 111: I think it's more efficient to just return the data that the client needs - even at the expense of some CPU work in the servlet. Also, querying METADATA without the body is probably more efficient on the Gmail side, and would be a win overall.

src/test/java/GmailActionableEmailsServletTest.java
```diff
@@ -0,0 +1,166 @@
...
+  @Test
+  public void validResponse() throws Exception {
+    Mockito.when(request.getParameter("subjectLinePhrases"))
+        .thenReturn(SUBJECT_LINE_PHRASES_STRING);
+    Mockito.when(request.getParameter("unreadOnly")).thenReturn(String.valueOf(true));
+    Mockito.when(request.getParameter("nDays")).thenReturn(String.valueOf(DEFAULT_N_DAYS));
+
+    // List passed to getActionableEmails should be words from query string
+    // parsed by comma separator
+    List<String> listOfSubjectWords = Arrays.asList(SUBJECT_LINE_PHRASES_STRING.split(","));
```
antmar, line 153: I suggest defining a separate constant for the expected list. That way you are testing just the inputs/outputs instead of duplicating the parsing logic in the test.

src/main/java/com/google/sps/servlets/GmailActionableEmailsServlet.java
```diff
@@ -0,0 +1,133 @@
...
+
+    boolean unreadOnly;
+    String unreadOnlyParameter = request.getParameter("unreadOnly");
+    if (unreadOnlyParameter == null) {
+      response.sendError(400, "unreadOnly must be present in request");
+      return;
+    }
+    if (!unreadOnlyParameter.equals("false") && !unreadOnlyParameter.equals("true")) {
+      response.sendError(400, "unreadOnly must be a boolean");
+    }
```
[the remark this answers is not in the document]
austinteshuba: Okay yes totally fair. I like the idea of the optional parameter. Thanks!

Also, to follow up here, this actually was tested using the methods we've used previously, but I think this method might actually be failing. The test checks if the error code is 400, which in this case it is. However, there is also a response, since the request continues processing erroneously. I'm going to file an issue (#128) to look at this in a separate PR (and ensure that only an error is returned, and nothing else, when an error is sent in a servlet)

src/main/java/com/google/sps/servlets/GmailActionableEmailsServlet.java
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * Parses a list of values (contained as a single parameter) from a request. Requested parameter
+   * should have format: ..."parameter=value1,value2,value3"...
+   *
+   * @param request HttpServletRequest containing parameter with above format
+   * @param parameter name of the request parameter
+   * @return parsed list of request parameter values. Empty list if parameter is empty
+   */
+  private List<String> getListFromQueryString(HttpServletRequest request, String parameter)
```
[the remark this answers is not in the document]
austinteshuba: Agreed, moving!

src/test/java/GmailActionableEmailsServletTest.java
austinteshuba: Sounds good to me!

src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -96,6 +96,25 @@ static String combineSearchQueries(String... queries) {
...
 
+  /**
+   * Get list of actionable emails that meet specified criteria
+   *
+   * @param messageFormat GmailClient.MessageFormat setting to control how much of each message is
+   *     returned
+   * @param subjectLinePhrases list of words that gmail should look for in the subject line. Emails
+   *     will be returned as long as one of the passed phrases are present.
+   * @param unreadOnly true if emails must be unread, false otherwise
+   * @param nDays how many days of emails Gmail should look through.
```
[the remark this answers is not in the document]
austinteshuba: Agreed - specifying now!

src/main/java/com/google/sps/model/GmailClientImpl.java
```diff
@@ -69,8 +69,38 @@ public Message getUserMessage(String messageId, MessageFormat format) throws IOE
...
+   * Lists out messages, but maps each user message to a specific message format
+   *
+   * @param messageFormat GmailClient.MessageFormat setting that specifies how much information from
+   *     each email to retrieve
+   * @param searchQuery search query to filter which results are returned (see:
+   *     https://support.google.com/mail/answer/7190?hl=en)
+   * @return list of messages with requested information
+   * @throws IOException if there is an issue with the GmailService
+   */
+  private List<Message> listUserMessagesWithFormat(
```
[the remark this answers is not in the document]
austinteshuba: Thanks!

src/main/java/com/google/sps/servlets/GmailActionableEmailsServlet.java
```diff
@@ -0,0 +1,133 @@
...
+
+    List<Message> actionableEmails =
+        gmailClient.getActionableEmails(
+            GmailClient.MessageFormat.FULL, subjectLinePhrases, unreadOnly, nDays);
+    JsonUtility.sendJson(response, actionableEmails);
+  }
+
+  /**
+   * Parses a list of values (contained as a single parameter) from a request. Requested parameter
+   * should have format: ..."parameter=value1,value2,value3"...
```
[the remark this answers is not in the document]
austinteshuba: Talked offline - leaving as is since Gmail (which is currently the only user of this utility) does not consider punctuation for search. 

If this is needed in the future, it will be developed and handled then.

src/main/java/com/google/sps/servlets/GmailActionableEmailsServlet.java
austinteshuba: Agreed- I made some changes to the Gmail client to facilitate this, as well as created a new POJO to ensure that only the correct data is being returned. 

Please let me know what you think - this was a substantial change.

src/main/java/com/google/sps/model/GmailClientImpl.java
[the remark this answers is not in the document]
austinteshuba: Thanks!

src/main/java/com/google/sps/model/GmailClient.java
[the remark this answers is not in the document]
austinteshuba: Agreed - done! Thanks

src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -42,6 +45,18 @@
...
+  /**
+   * Gets a message from a user's Gmail account. Format assumed to be METADATA. Returned message
+   * will only include email message ID, labels, and specified headers
+   *
+   * @param messageId messageID (retrieved from listUserMessages) of the desired Message
+   * @param metadataHeaders list of names of headers (e.g. "From") that should be included
+   * @return a Message object with the requested information
+   * @throws IOException if an issue occurs with the Gmail service
+   */
+  Message getUserMessageWithMetadataHeaders(String messageId, List<String> metadataHeaders)
```
antmar, line 57: This seems like an implementation details of `GmailClientImpl`. Does it need to be exposed via the interface?

src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -189,4 +202,30 @@ static String oneOfPhrasesInSubjectLineQuery(List<String> phrases) {
...
+   * Given a list of message headers, extract a single header with the specified name
+   *
+   * @param message a Gmail message object. Must be METADATA or FULL format
+   * @param headerName name of header that should be extracted
+   * @return first header in list with name headerName (this method does not handle duplicate
+   *     headers)
+   * @throws GmailMessageFormatException if the message does not contain the specified header (and
+   *     is thus the wrong format)
+   */
+  static MessagePartHeader extractHeader(Message message, String headerName) {
```
antmar, line 216: Perhaps move static helper to a separate static utility class

src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -189,4 +202,30 @@ static String oneOfPhrasesInSubjectLineQuery(List<String> phrases) {
...
+    MessagePart payload = message.getPayload();
+    if (payload == null) {
+      throw new GmailMessageFormatException("No headers present!");
```
antmar, line 219: It's also possible that payload.getHeaders() is empty, but you would return a different error. Perhaps make this error message more specific.

src/test/java/GmailActionableEmailsServletTest.java
```diff
@@ -37,16 +41,42 @@
 
   private static final Gson gson = new Gson();
 
-  // Spaces should be parsed correctly by request.getParameter if query string is properly encoded
-  private static final String SUBJECT_LINE_PHRASES_STRING = "Action Word One, ActionWordTwo";
+  // Spaces within words are fine (and would be encoded/decoded by request.getParameter)
+  private static final String SUBJECT_LINE_PHRASES_STRING = "\"Action Word One\",\"ActionWordTwo\"";
+  private static final List<String> SUBJECT_LINE_PHRASES_LIST =
+      Arrays.asList("\"Action Word One\"", "\"ActionWordTwo\"");
```
antmar, line 47: Are you assuming that the client will always add quotes around the phrases? Does it even matter here?

I think you are handling the quotes in the GmailClientImpl (ignoring/adding them). Maybe just assume the inputs don't contain quotes, and don't send quotes in your actual client. It just complicates things and makes it seem like the quotes are significant. (If the user enters quotes, then they'll be ignored, and that's fine.)

src/main/java/com/google/sps/model/GmailClient.java
austinteshuba: This actually is a supported get operation from the Gmail API (where metadataHeaders can simply be passed as a parameter), so I think this belongs here semantically.

I'm hoping that others also use this, since it is more efficient than the standard METADATA format if you don't need all the headers.

src/main/java/com/google/sps/model/GmailClient.java
austinteshuba: Done!

src/main/java/com/google/sps/model/GmailClient.java
austinteshuba: Agreed - fixed!

src/test/java/GmailActionableEmailsServletTest.java
austinteshuba: Yes fair enough - I feel extra safe making this assumption here since the quotes handling is done by the GmailClient (and it is tested). 

Changing!

PR #113, opened 2020-07-24
Remove GmailTestBase

austinteshuba:
This is a follow up to PR #106. I removed the GmailTestBase due to the interface anti-pattern than @gbuenoandrade mentioned, but I forgot to push before merging. As a result, those changes were added to master. 

This reverses that addition.


src/test/java/GmailResponseHelperImplTest.java
```diff
@@ -12,19 +12,150 @@
...
 
 /**
  * Tests the GmailResponseHelperImpl and ensures that statistics for the GmailResponse are
  * calculated correctly
  */
-public class GmailResponseHelperImplTest implements GmailTestBase {
+public class GmailResponseHelperImplTest {
   private static final GmailResponseHelper gmailResponseHelper = new GmailResponseHelperImpl();
 
+  int DEFAULT_N_DAYS = 7;
```
gbuenoandrade, line 36: You forgot to prefix the constants with `static final`

austinteshuba: Thanks! Fixed

austinteshuba:
I decided to avoid using interfaces for constants, since Effective Java (the guide we agreed offline to mostly follow) discourages this.

PR #116, opened 2020-07-27
Implement Places Client

tttzach:
Implement Places Client to search for locations nearby for the purpose of changing a generic place type to an exact address in the Go panel.


src/main/java/com/google/sps/exceptions/PlacesException.java
```diff
@@ -0,0 +1,25 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.exceptions;
+
+public class PlacesException extends Exception {
```
austinteshuba, line 17: nit: Javadoc on the class and constructors

austinteshuba: Also wondering under what circumstances a PlacesException would be thrown - but I think that would/will be explained by the javadoc so feel free not to reply to this comment

src/main/java/com/google/sps/model/PlacesClientImpl.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+
+  /**
+   * Gets all formatted addresses from given response. Scope of method is public for testing
+   * purposes.
+   *
+   * @param result The PlacesSeachResponse object to get formatted addresses from
+   */
+  public static List<String> getFormattedAddresses(PlacesSearchResponse response) {
```
austinteshuba, line 54: I would move this to the PlacesClient.java interface (which can contain static methods), since this pertains to the overall function of any PlacesClient, not just the impl

austinteshuba: Also javadoc should include some information about what the format of a formatted address is

src/main/java/com/google/sps/exceptions/PlacesException.java
tttzach: Good call on the javadoc comment. This exception is just to keep things consistent. For example, there's a DirectionsException for all things wrong with the Directions API. Naturally, I think it makes sense to have something identical for the Places API. Thanks!

src/main/java/com/google/sps/model/PlacesClientImpl.java
tttzach: Made the changes! Let me know if it's better now.

src/main/java/com/google/sps/exceptions/PlacesException.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/PlacesClientImpl.java
tttzach: `RESOLVED`

PR #118, opened 2020-07-27
Implement Locations Utility class

tttzach:
Implement logic to parse for locations in Tasks for the purpose of the Go magic feature.


src/main/java/com/google/sps/model/TasksClientImpl.java
```diff
@@ -61,6 +62,17 @@ private TasksClientImpl(Credential credential) {
     return taskLists != null ? taskLists : new ArrayList<>();
   }
 
+  @Override
+  public List<TaskList> listTaskLists(List<String> taskListTitles) throws IOException {
+    // returns null if no taskLists exist. Convert to empty list for ease.
+    List<TaskList> taskLists = tasksService.tasklists().list().execute().getItems();
+    taskLists = taskLists != null ? taskLists : new ArrayList<>();
+    taskLists.stream()
```
austinteshuba, line 70: This filtering logic should likely be tested (and as of now it cannot be tested) - I would actually suggest not making an overload for this method and instead using the more general `listTaskLists()` method, then creating a static helper method to do this filtering

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -61,33 +73,78 @@ public GoServlet() throws IOException {
...
+  public List<String> getLocations(String prefix, List<Task> tasks) {
+    return tasks.stream()
+        .map(task -> task.getNotes())
+        .filter(Objects::nonNull)
+        .map(notes -> getLocation(prefix, notes))
+        .filter(place -> !place.equals("No " + prefix))
+        .collect(Collectors.toList());
+  }
+
+  private String getLocation(String prefix, String taskNotes) {
```
austinteshuba, line 115: While not required, some comments here would be nice to understand what the regex is filtering for

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -61,33 +73,78 @@ public GoServlet() throws IOException {
...
+
+  /**
+   * Parses for locations in Tasks. Scope is public for testing purposes.
+   *
+   * @param prefix String which represents the prefix wrapped in square brackets to look for. (e.g.
+   *     Location if looking for [Location: ])
+   * @param tasks List of tasks to parse for locations from.
+   * @return List of strings representing the locations.
+   */
+  public List<String> getLocations(String prefix, List<Task> tasks) {
```
austinteshuba, line 106: If this logic is separate enough to be tested in isolation as a public method, it should probably be in a client as a static method.

src/test/java/GoServletTest.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.List;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.Mockito;
+
+@RunWith(JUnit4.class)
+public class GoServletTest extends AuthenticatedServletTestBase {
```
austinteshuba, line 29: The servlet test should be testing that the response is correctly created given some input from the client. 

Currently, while this does a good job at testing the `getLocations` method, this does not do this. You likely need more tests here for the servlet responses themselves.

tttzach: You're absolutely right about this. I think the approach I went with to make sure the Go servlet PR isn't overly large does not take testing into consideration as much as it should. Now that I think of it - it's probably better to just merge this PR with one that passes the addresses into Directions and returns the result to the client so that on top of being able to test the servlet as a whole, a working MVP is available as well. The Places API can be an "add-on" for later.

tttzach:
Note to self - implement working MVP here. i.e. Pass locations to Directions and display on client. The end result of this PR being a Go panel which takes addresses from **ALL** tasks and displays the recommended order of travel on the client.

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: Good call, regex can be confusing on its own.

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: I agree that it should probably be somewhere else, I have moved it into a Locations Utility class since I don't think it fits into any of the existing clients and that it requires any dependency injection like the other clients we have. Thanks for pointing that out!

src/test/java/GoServletTest.java
tttzach: This doesn't apply anymore because I have moved the `getLocations` logic into a utility class of its own after addressing the comment directly above this. Thanks!

tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/TasksClientImpl.java
tttzach: That's fair - after addressing the comments below, this PR is now more of a Locations Utility PR so I will definitely keep that in mind when polishing up the Go servlet in another PR. Thanks!

tttzach: `RESOLVED`

src/main/java/com/google/sps/utility/LocationsUtility.java
```diff
@@ -0,0 +1,61 @@
...
+package com.google.sps.utility;
+
+import com.google.api.services.tasks.model.Task;
+import java.util.List;
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class LocationsUtility {
```
antmar, line 24: `public static class`? You are not instantiating this I assume.

src/main/java/com/google/sps/utility/LocationsUtility.java
```diff
@@ -0,0 +1,61 @@
...
+        .filter(place -> !place.equals("No " + prefix))
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Parses for string enclosed in [prefix: ] in notes of a task
+   *
+   * @param prefix String to look for in the square brackets
+   * @param taskNotes String to parse from, usually in the form "... [prefix: ... ] ..."
+   * @return String enclosed in [prefix: ] or No prefix if no enclosure found
```
antmar, line 47: It's undesirable to return "No (prefix)" in the not-found case, when any string is also a valid return value on success.

If you don't care about the difference between "prefix not found" and "the location is empty", then just return an empty string. If you care about the difference, then return a null string (and clearly document it).

src/main/java/com/google/sps/utility/LocationsUtility.java
```diff
@@ -0,0 +1,61 @@
...
+   */
+  private static String getLocation(String prefix, String taskNotes) {
+    // Regular expression matches the characters [prefix: and ] literally, (.*?) signifies the 1st
+    // capturing group which matches any character except for line terminators
+    String regex = "\\[" + prefix + ": (.*?)\\]";
+    Pattern pattern = Pattern.compile(regex);
```
antmar, line 53: You should compile the regex in `getLocations` and pass it to this method. Doing it for each task is expensive.

src/test/java/LocationsUtilityTest.java
```diff
@@ -0,0 +1,121 @@
...
+  private static final List<Task> TASKS_WITH_NO_NOTES = ImmutableList.of(TASK_WITH_NO_NOTES);
+  private static final List<Task> TASKS_WITH_NO_LOCATION = ImmutableList.of(TASK_WITH_NO_LOCATION);
+  private static final List<Task> TASKS_WITH_ONE_LOCATION =
+      ImmutableList.of(TASK_WITH_ONE_LOCATION);
+  private static final List<Task> TASKS_WITH_TWO_LOCATIONS =
+      ImmutableList.of(TASK_WITH_TWO_LOCATIONS);
+  private static final List<Task> TASKS_WITH_ONE_LOCATION_ENCLOSED_INCORRECTLY =
+      ImmutableList.of(TASK_WITH_ONE_LOCATION_ENCLOSED_INCORRECTLY);
+  private static final List<Task> TASKS_WITH_EMPTY_LOCATION =
+      ImmutableList.of(TASK_WITH_EMPTY_LOCATION);
```
antmar, line 49: I don't think these constants are necessary. You could just create the `ImmutableList` when you need it. You are already doing that for the expected results in all your tests.

However, if you want to keep these as constants, then make it clear there is one task in each list: `ONE_TASK_WITH_NO_NOTES`, etc.

src/test/java/LocationsUtilityTest.java
```diff
@@ -0,0 +1,121 @@
...
+    // Obtain location in the task notes of one task with no location.
+    Assert.assertEquals(
+        ImmutableList.of(), LocationsUtility.getLocations(PREFIX, TASKS_WITH_NO_LOCATION));
+  }
+
+  @Test
+  public void getOneLocation() {
+    // Obtain location in the task notes of one task with one location.
+    Assert.assertEquals(
+        ImmutableList.of("Google Kitchener"),
```
antmar, line 83: The expected location should be a constant, so that it's defined somewhere close to where you define the test inputs.

src/test/java/LocationsUtilityTest.java
```diff
@@ -0,0 +1,121 @@
...
+        ImmutableList.of(),
+        LocationsUtility.getLocations(PREFIX, TASKS_WITH_ONE_LOCATION_ENCLOSED_INCORRECTLY));
+  }
+
+  @Test
+  public void getEmptyLocation() {
+    // Obtain location in the task notes of one task with one location with [Location: ] tag but
+    // nothing inside of it.
+    Assert.assertEquals(
+        ImmutableList.of(""), LocationsUtility.getLocations(PREFIX, TASKS_WITH_EMPTY_LOCATION));
```
antmar, line 110: Is empty location really a valid location? Or is it a special case of a missing/malformed location?

(Maybe there is a valid use case for empty locations - it depends on how this utility will be used.)

src/main/java/com/google/sps/utility/LocationsUtility.java
```diff
@@ -0,0 +1,61 @@
...
+    return tasks.stream()
+        .map(task -> task.getNotes())
+        .filter(Objects::nonNull)
+        .map(notes -> getLocation(prefix, notes))
+        .filter(place -> !place.equals("No " + prefix))
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Parses for string enclosed in [prefix: ] in notes of a task
```
antmar, line 43: Document how multiple location tags are handled.

src/test/java/LocationsUtilityTest.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @Test
+  public void getLocationAllTasks() {
+    // Obtain location in the task notes of five tasks each with either one, empty or no location as
+    // noted in the individual tests above.
+    Assert.assertEquals(
+        ImmutableList.of("Google Kitchener", "Google Kitchener", ""),
+        LocationsUtility.getLocations(PREFIX, ALL_TASKS));
```
antmar, line 119: Would it make sense to remove duplicate locations in the output? I have no strong opinion either way - it just seems like you are going to have to do this deduplication anyways.

src/main/java/com/google/sps/utility/LocationsUtility.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.api.services.tasks.model.Task;
+import java.util.List;
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class LocationsUtility {
+  /**
+   * Parses for locations in Tasks.
```
antmar, line 26: Document how missing, empty, and duplicate locations are handled.

src/main/java/com/google/sps/utility/LocationsUtility.java
tttzach: Yep you're right - thanks! I changed it to `public final class` and made the constructor private.

src/main/java/com/google/sps/utility/LocationsUtility.java
tttzach: I agree, it was just for filtering purposes. I have changed it to return an empty string instead - this is a lot better because then the method filters out empty locations as well.

src/main/java/com/google/sps/utility/LocationsUtility.java
tttzach: Fair point! I wasn't aware how expensive regex is. Changed it, thanks!

src/test/java/LocationsUtilityTest.java
tttzach: Agreed, the string "Google Kitchener" is also scattered throughout the file so it makes sense to be a constant. Thanks!

src/test/java/LocationsUtilityTest.java
tttzach: No it is not a valid location, the utility's job is to just parse and the validation should be done in the servlet but since the "No prefix" is changed to "", these are now filtered as well. Thanks for pointing it out!

src/main/java/com/google/sps/utility/LocationsUtility.java
tttzach: Done, thanks!

src/test/java/LocationsUtilityTest.java
tttzach: Yea I can see why it makes sense to deduplicate in the `getLocations` method itself plus it seems like adding `.distinct()` to `.stream()` did the deduplication perfectly. Thanks for the suggestion!

src/main/java/com/google/sps/utility/LocationsUtility.java
tttzach: Added, thanks!

src/main/java/com/google/sps/utility/LocationsUtility.java
antmar: Update the `@return` comment line with the new behavior.

src/test/java/LocationsUtilityTest.java
```diff
@@ -99,23 +82,33 @@ public void getIncorrectlyEnclosedLocation() {
...
     Assert.assertEquals(
-        ImmutableList.of(""), LocationsUtility.getLocations(PREFIX, TASKS_WITH_EMPTY_LOCATION));
+        ImmutableList.of(),
+        LocationsUtility.getLocations(PREFIX, ImmutableList.of(TASK_WITH_EMPTY_LOCATION)));
   }
 
   @Test
   public void getLocationAllTasks() {
     // Obtain location in the task notes of five tasks each with either one, empty or no location as
     // noted in the individual tests above.
```
antmar: It's not clear what is the point of this test. You've already tested the tasks that produce no location, and the tasks that produce `LOCATION`. I think this test case would make more sense if you had two tasks that produced different locations, and you check that both are returned. 

(You can keep the other tasks which produce duplicate locations or no locations at all. I am specifically pointing out the need to test returning multiple locations.)

tttzach: Fair point! Mindset starting to change now that there's so many pending PRs and so little time. I will be more careful.

src/main/java/com/google/sps/utility/LocationsUtility.java
tttzach: Thanks for pointing this out, i'll be more careful in the future changes.

src/test/java/LocationsUtilityTest.java
tttzach: Fair point! I will be more careful with changing other stuff based on review changes. Thanks for pointing it out!

PR #119, opened 2020-07-27
Create simple client side implementation of Assign Servlet

austinteshuba:
Currently gets the response from the server for hardcoded parameters. Checks for "Action Requested" or "Action Required" in subject line.

Emails:
Screenshot 2020-07-27 at 5 52 44 PM

Result: 
Screenshot 2020-07-27 at 5 52 59 PM

Dependent on PR #112


src/main/webapp/utilities.js
```diff
@@ -90,3 +90,23 @@ function getDateInLocalTimeZone(dateObject = new Date()) {
...
+ * Given a list of words, encode them for a query string with commas separating
+ * the values.
+ *
+ * @param {string[]} listOfValues list of words that should be encoded
+ * @param {boolean} surroundWithQuotes true if words should be surrounded with
+ *     quotes, false if the words can be encoded as is
+ * @return {string} the value for a query parameter, with each value seperated
+ *     by a comma.
+ */
+function encodeListForUrl(listOfValues, surroundWithQuotes) {
```
tttzach, line 104: Seems like this function is called once with `surroundWithQuotes = true`, would there be a case where this is used with `surroundWithQuotes = false`?

src/main/webapp/populateData.js
```diff
@@ -311,3 +312,60 @@ function populateGo() {
...
+ */
+function setUpAssign() {
+  const assignContent = document.querySelector('#assign');
+
+  const subjectLinePhrases = ['Action Required', 'Action Requested'];
+  fetchActionableEmails(subjectLinePhrases, true, 7)
+      .then((response) => {
+        assignContent.innerText = response
+            .map((obj) => obj.snippet)
+            .reduce((a, b) => a + '\n' + b);
```
tttzach, line 327: Map and reduce makes your code look clean - I should probably use them more!

src/main/webapp/populateData.js
```diff
@@ -311,3 +312,60 @@ function populateGo() {
...
+
+  return fetch(queryString)
+      .then((response) => {
+        switch (response.status) {
+          case 200:
+            return response.json();
+          case 403:
+            throw new AuthenticationError();
+          default:
+            throw new Error(response.status + ' ' + response.statusText);
```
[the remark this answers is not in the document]
austinteshuba: I can, but this is used elsewhere in the script file. For consistency, I think this should stay the same for now. 

Plus this is a rather simple string concatenation, so it probably isn't prohibited under style guide (it only talks about [complicated string concatenation being less preferred than template literals](https://google.github.io/styleguide/jsguide.html#features-strings-template-strings).

src/main/webapp/populateData.js
austinteshuba: They're a lot of fun to use once you get the hang of them!

src/main/webapp/utilities.js
austinteshuba: Good catch - probably not. If for some reason in the future this does arise, it can be developed then. I'm removing this :)

austinteshuba: This isn't really for me, but for others - if anyone else wants to make a list but the words don't need to be surrounded by quotes (i.e. `<some url>?list=phrase one,phrase two` is fine instead of `<some url>?list="phrase one","phrase two"`) they can still use this utility that way

src/main/webapp/populateData.js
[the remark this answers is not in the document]
tttzach: Okay sure, it's definitely not an important change rather it was just in case you wanted to do the same as in line 357. Thanks for addressing my comment!

src/main/webapp/utilities.js
tttzach: Hmm okay, unless there is some use case where it is a must to have no quotes around the words, it might be better to just stick with the one with quotes throughout the entire codebase. Though this thought is only because I can't think of a use case, let me know if there is!

src/main/webapp/populateData.js
tttzach: Wasn't a fan of .stream() but it's starting to grow on me not gonna lie

gbuenoandrade, line 327: Nit: I think you can use `.join("\n")` instead of the "manual" reduce

src/main/webapp/utilities.js
```diff
@@ -90,3 +90,23 @@ function getDateInLocalTimeZone(dateObject = new Date()) {
...
+  return listOfValues.map((value) => {
+    if (surroundWithQuotes) {
+      // Will add quotes, if not already present
+      value = `"${value.replace(/^"(.*)"$/, '$1')}"`;
+    }
+    return encodeURI(value);
+  }).reduce((a, b) => a + ',' + b);
```
gbuenoandrade, line 111: Nit: `.join()`

src/main/webapp/utilities.js
gbuenoandrade: +1 to always using quotes

src/main/webapp/utilities.js
```diff
@@ -90,3 +90,23 @@ function getDateInLocalTimeZone(dateObject = new Date()) {
...
+ * @param {string[]} listOfValues list of words that should be encoded
+ * @param {boolean} surroundWithQuotes true if words should be surrounded with
+ *     quotes, false if the words can be encoded as is
+ * @return {string} the value for a query parameter, with each value separated
+ *     by a comma.
+ */
+function encodeListForUrl(listOfValues, surroundWithQuotes) {
+  return listOfValues.map((value) => {
+    if (surroundWithQuotes) {
+      // Will add quotes, if not already present
```
gbuenoandrade, line 107: For simplicity, I think it should always surround the elements with quotes when `surroundWithQuotes = true`. Also, that's what your API promises.

src/main/webapp/utilities.js
austinteshuba: No worries, changing over. 

If this is needed in the future, it's easy enough to implement :)

src/main/webapp/populateData.js
austinteshuba: Cool - wasn't familiar with `.join()`. Thanks!

src/main/webapp/utilities.js
austinteshuba: Sounds good to me - thanks!

src/main/webapp/utilities.js
```diff
@@ -90,3 +90,15 @@ function getDateInLocalTimeZone(dateObject = new Date()) {
...
+/**
+ * Given a list of phrases, encode them for a query string with commas
+ * separating the values and quotes surrounding the phrases
+ *
+ * @param {string[]} listOfValues list of words that should be encoded.
+ * @return {string} the value for a query parameter, with each value separated
+ *     by a comma.
+ */
+function encodeListForUrl(listOfValues) {
+  return listOfValues.map((value) => encodeURI(`"${value}"`)).join();
```
antmar, line 103: What's the value in putting quotes around the phrases? There is no value on the client, and they are ignored by the servlet.

src/main/webapp/populateData.js
```diff
@@ -311,3 +312,60 @@ function populateGo() {
...
 
+/**
+ * Set up the assign panel. For now, this just prints the response
+ * from the server for /gmail-actionable-emails
+ */
+function setUpAssign() {
+  const assignContent = document.querySelector('#assign');
+
+  const subjectLinePhrases = ['Action Required', 'Action Requested'];
+  fetchActionableEmails(subjectLinePhrases, true, 7)
```
antmar, line 323: Perhaps define constants for the other parameters here, so that it's easier to understand which emails are being fetched.

src/main/webapp/utilities.js
austinteshuba: Good point - no need for this. Thanks!

src/main/webapp/populateData.js
austinteshuba: Good idea - thanks!

austinteshuba:
@austinteshuba don't merge until PR #112 is approved.

PR #120, opened 2020-07-28
Comment out postAndGetTaskList

tttzach:
Commenting the JavaScript function postAndGetTaskList out since it adds tasks to user's Tasks account on every run.


austinteshuba:
^^ Should be fine - was there for testing purposes / proof that this worked in the case of demo. LGTM

Issue #68, opened 2020-07-13
Parallelize/Refactor the populatePanel functions


austinteshuba:
Make the populate functions operate asynchronously, and abstract them a bit so they're less repetitive

austinteshuba:
This is a non-issue, as they are already async

PR #122, opened 2020-07-29
Make populate functions asynchronous

austinteshuba:
Closes Issue #68. 

All the panels now load at the same time, reducing lag time. 

Changes to the linter were made so it would recognize the keyword `async`. This also required additional rules to be clarified, namely with trailing commas.


.eslintrc.yml
```diff
@@ -2,3 +2,15 @@ extends: ["eslint:recommended", "google"]
...
+}
+"rules": {
+  "comma-dangle": ["error", {
+    "arrays": "always",
+    "objects": "always",
+    "imports": "never",
+    "exports": "never",
+    "functions": "never"
+  }]
+}
```
gbuenoandrade, line 16: Use YAML syntax instead of JSON (or refactor the whole file and change the extension)

.eslintrc.yml
```diff
@@ -2,3 +2,15 @@ extends: ["eslint:recommended", "google"]
 env:
   es6: true
   browser: true
+"parserOptions": {
+  "ecmaVersion": 2017
```
gbuenoandrade, line 6: Does this make setting `es6` to true obsolete?

austinteshuba:
I don't want to change the language level, so I'm going to redo this with promises.

PR #124, opened 2020-07-29
Use promises to populate panels asynchronously 

austinteshuba:
Closes issue #68

### DIFF
--- a/src/main/webapp/auth.js
+++ b/src/main/webapp/auth.js
@@ -56,16 +56,16 @@ function handleAuthenticationState() {
     signInButton.setAttribute('hidden', '');
     featureContainer.removeAttribute('hidden');
 
-    // Populate information panels at top of dashboard
-    populateGmail();
-    populateTasks();
-    populateCalendar();
+    // Populate information panels at top of dashboard asynchronously
+    new Promise(() => populateGmail());
+    new Promise(() => populateTasks());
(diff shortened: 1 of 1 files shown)


tttzach:
LGTM - i'll be sure to do the same for the future JavaScript function calls.

src/main/webapp/auth.js
```diff
@@ -55,16 +55,16 @@ function handleAuthenticationState() {
...
     featureContainer.removeAttribute('hidden');
 
-    // Populate information panels at top of dashboard
-    populateGmail();
-    populateTasks();
-    populateCalendar();
+    // Populate information panels at top of dashboard asynchronously
+    new Promise(() => populateGmail());
+    new Promise(() => populateTasks());
+    new Promise(() => populateCalendar());
```
gbuenoandrade, line 61: Nit: I'm not super familiar with asynchronous programming in JavaScript, but it looks a bit weird to just create those promises like that...

Why have you decided not to make the `populate.*()` methods async?

antmar: `populate.*()` methods are already "async" in that the calls to `fetch` return a promise and don't block.

austinteshuba:
Closing PR as no changes are needed to achieve this - The promises already remove the blockers, and allow the main thread to continue working.

PR #125, opened 2020-07-29
UI for the settings of the Assign Panel

austinteshuba:
Create settings panel for Assign panel, and tighten up the UI. 

This PR focuses on elements that use JavaScript (lots of buttons, creating/deleting entries, etc). 

Settings Panel (toggled on and off by pressing gear in upper-right corner):
Screenshot 2020-07-30 at 10 34 45 PM


src/main/webapp/panelStyle.js
```diff
@@ -0,0 +1,174 @@
...
+/**
+ * Shows/hides the settings display in the assign panel
+ *
+ * @param {Event} event event generated from click
+ */
+function displayAssignSettings(event) {
+  // Change icon appearance
+  toggleIconHandler(event);
+
+  // TODO: show/hide the settings display
```
tttzach, line 63: Is this the same as the one in assignPanelStyle.js?

austinteshuba: Fixed!

src/main/webapp/panelStyle.js
```diff
@@ -0,0 +1,168 @@
...
+  const currentIcon = event.target;
+  const currentDiv = currentIcon.parentNode;
+  const newIcon = currentDiv.querySelector('img[hidden]');
+
+  currentIcon.setAttribute('hidden', '');
+  newIcon.removeAttribute('hidden');
+}
+
+/**
+ * Will increase a the integer value of a text element up to a
```
gbuenoandrade, line 55: nit: "increase a the integer"

src/main/webapp/panelStyle.js
```diff
@@ -0,0 +1,168 @@
...
+ */
+function incrementElement(elementId, max = 30) {
+  const targetElement = document.getElementById(elementId);
+  const currentValue = parseInt(targetElement.innerText);
+
+  targetElement.innerText = Math.min(max, currentValue + 1);
+}
+
+/**
+ * Will decrease a the integer value of a text element down to a
```
gbuenoandrade, line 70: nit: "decrease a the integer"

src/main/webapp/panelStyle.js
```diff
@@ -0,0 +1,168 @@
...
+ * remove button (to delete the entry)
+ *
+ * @param {string} inputElementId the id of the input field
+ * @param {string} listElementId the id of the list that contains the entries
+ */
+function createTextListElement(inputElementId, listElementId) {
+  const inputElement = document.getElementById(inputElementId);
+
+  // Remove commas from phrase, since this is currently the delimiter
+  // for query strings and they are not recognized by most Google services
```
gbuenoandrade, line 95: What do you mean by "not recognized by most Google services"? You can't use commas even if you escape them?

src/main/webapp/panelStyle.js
```diff
@@ -0,0 +1,168 @@
...
+  // TODO: Nicer client-side validation (tell user why this isn't permitted)
+  // (Issue #135)
+  let phrase = inputElement.value;
+  if (phrase === null || phrase.length === 0) {
```
gbuenoandrade, line 99: Maybe also return (or do something else) if the phrase is already part of `list`?

src/main/webapp/panelStyle.js
```diff
@@ -0,0 +1,168 @@
...
+    return;
+  }
+  phrase = phrase.split(',').join('');
```
gbuenoandrade, line 102: nit: `phrase.replace(',', '')`

src/main/webapp/panelStyle.js
austinteshuba: Oops - fixed!

src/main/webapp/panelStyle.js
austinteshuba: Oops - fixed

src/main/webapp/panelStyle.js
austinteshuba: I should clarify this in the comment. 

You **can** use them for visual purposes (i.e. naming a task with a comma, or having a tasklist that contains a comma in the title, or having a task notes with a comma). But in terms of search queries (for Gmail and otherwise), punctuation is ignored.

src/main/webapp/panelStyle.js
austinteshuba: Agreed! Added - duplicates are rejected and the input is now cleared in the case of a duplicate entry

src/main/webapp/panelStyle.js
austinteshuba: Thanks - fixed!

src/main/webapp/panelStyle.js
gbuenoandrade: ACK

PR #129, opened 2020-07-30
Implementation of Get request for the plan-mail feature

AymarMuhikira:
Implemented the function to get and display the potential meeting times to read emails.

Screenshot 2020-07-30 at 2 37 54 PM - Display 1


src/main/java/com/google/sps/data/PlanMailResponse.java
```diff
@@ -0,0 +1,59 @@
...
+
+package com.google.sps.data;
+
+import com.google.sps.utility.DatePair;
+import java.util.*;
+
+/** Class containing the response to be converted to Json. */
+public final class PlanMailResponse {
+
+  final int wordCount;
```
austinteshuba, line 23: nit: use private

src/main/java/com/google/sps/data/PlanMailResponse.java
```diff
@@ -0,0 +1,59 @@
...
+      int minutesToRead,
+      List<DatePair> potentialMeetingTimes) {
+    this.wordCount = wordCount;
+    this.averageReadingSpeed = averageReadingSpeed;
+    this.minutesToRead = minutesToRead;
+    this.potentialMeetingTimes = potentialMeetingTimes;
+  }
+
+  public int getWordCount() {
+    // Useful for testing purposes
```
austinteshuba, line 41: nit: Remove comment - they're just getter functions

src/main/java/com/google/sps/model/AuthenticationVerifier.java
```diff
@@ -14,15 +14,20 @@
 
 package com.google.sps.model;
 
+import com.google.sps.utility.KeyProvider;
 import java.io.IOException;
 import java.security.GeneralSecurityException;
 
 /** Contract for verifying user with Google Sign in API */
 public interface AuthenticationVerifier {
```
austinteshuba, line 22: Just a HU that this is already in master, so you probably need to merge master into `mvp/calendar-panel`, and then merge `mvp/calendar-panel` into this branch

src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -42,6 +42,18 @@
    */
   Message getUserMessage(String messageId, MessageFormat format) throws IOException;
 
+  // /**
```
austinteshuba, line 45: Remove unused code

src/main/java/com/google/sps/model/GmailClientImpl.java
```diff
@@ -83,6 +83,96 @@ public Message getUserMessage(String messageId, MessageFormat format) throws IOE
         .collect(Collectors.toList());
   }
 
+  // @Override
```
austinteshuba, line 86: Is this code supposed to be unused? If so, remove

src/main/java/com/google/sps/servlets/PlanMailServlet.java
```diff
@@ -0,0 +1,217 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import javax.mail.MessagingException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** GET function responds JSON string containing potential events to read mail. */
+@WebServlet("/plan-mail")
+public class PlanMailServlet extends AuthenticatedHttpServlet {
```
austinteshuba, line 51: I would name this free-time servlet or something like that, since this is what the servlet does

src/main/java/com/google/sps/servlets/PlanMailServlet.java
```diff
@@ -0,0 +1,217 @@
...
+   */
+  @Override
+  public void doGet(
+      HttpServletRequest request, HttpServletResponse response, Credential googleCredential)
+      throws IOException {
+    assert googleCredential != null
+        : "Null credentials (i.e. unauthenticated requests) should already be handled";
+
+    CalendarClient calendarClient = calendarClientFactory.getCalendarClient(googleCredential);
+    GmailClient gmailClient = gmailClientFactory.getGmailClient(googleCredential);
```
austinteshuba, line 92: nit: do you need `gmailClient` here

src/main/java/com/google/sps/servlets/PlanMailServlet.java
```diff
@@ -0,0 +1,217 @@
...
+    long fiveDaysInMillis = TimeUnit.DAYS.toMillis(5);
+    Date timeMin = calendarClient.getCurrentTime();
+    Date timeMax = new Date(timeMin.getTime() + fiveDaysInMillis);
+    List<Event> calendarEvents = getEvents(calendarClient, timeMin, timeMax);
+
+    FreeTimeUtility freeTimeUtility = new FreeTimeUtility(timeMin);
+    for (Event event : calendarEvents) {
```
austinteshuba, line 99: This should probably be offloaded into a helper method for readability.

src/main/java/com/google/sps/servlets/PlanMailServlet.java
```diff
@@ -0,0 +1,217 @@
...
+      DateTime start = event.getStart().getDateTime();
+      start = start == null ? event.getStart().getDate() : start;
```
austinteshuba, line 101: I don't know if this handled the null case, since if `end==null`, and it's now set to `null`, `end.getValue()` will throw a null pointer

src/main/java/com/google/sps/servlets/PlanMailServlet.java
```diff
@@ -0,0 +1,217 @@
...
+      System.out.println(e);
+    }
+    for (Message message : unreadMessages) {
+      try {
+        wordCount += getMessageSize(message);
+      } catch (MessagingException | IOException e) {
+        System.out.println(e);
+      }
+    }
+    return 104;
```
austinteshuba, line 161: This does not return a real value

src/main/java/com/google/sps/servlets/PlanMailServlet.java
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  /**
+   * Get the events in the user's calendars
+   *
+   * @param calendarClient either a mock CalendarClient or a calendarClient with a valid credential
+   * @return List of Events from all of the user's calendars
+   * @throws IOException if an issue occurs in the method
+   */
+  private List<Event> getEvents(CalendarClient calendarClient) throws IOException {
```
austinteshuba, line 190: This is not currently used - delete if possible

src/main/webapp/populateData.js
```diff
@@ -329,3 +329,45 @@ function populateGo() {
...
+        const numWordsContainer = document.querySelector('#word-count');
+        numWordsContainer.innerText = planMailResponse.wordCount;
+        const averageReplyContainer = document.querySelector('#average-reply');
+        averageReplyContainer.innerText = planMailResponse.averageReadingSpeed;
+        const timeNeededContainer = document.querySelector('#time-needed');
+        timeNeededContainer.innerText = planMailResponse.minutesToRead;
+        const intervalContainer = document.querySelector('#free-interval');
+        intervalContainer.innerHTML = '';
+        for (const index in planMailResponse.potentialMeetingTimes) {
+          if (typeof index == 'string') {
```
austinteshuba, line 357: nit: I don't think this type check is needed

src/test/java/ClientIDServletTest.java
```diff
@@ -22,11 +21,13 @@
...
-  private static final ClientIDServlet servlet = new ClientIDServlet();
+
+  private static final String CLIENT_ID = "sampleValue";
+  private static final ClientIDServlet servlet = new ClientIDServlet(CLIENT_ID);
 
   @Test
   public void responseContainsClientId() throws Exception {
     servlet.doGet(request, response);
-    Assert.assertTrue(stringWriter.toString().contains(AuthenticationVerifier.CLIENT_ID));
+    Assert.assertTrue(stringWriter.toString().contains(CLIENT_ID));
```
tttzach, line 31: This seems to be one of my previous commits. Try rebasing and merging with master.

src/main/java/com/google/sps/data/PlanMailResponse.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/data/PlanMailResponse.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/model/AuthenticationVerifier.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/model/GmailClient.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/model/GmailClientImpl.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/PlanMailServlet.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/PlanMailServlet.java
AymarMuhikira: I moved the `gmailClient` to `getWordCount`. `Resolved`

src/main/java/com/google/sps/servlets/PlanMailServlet.java
AymarMuhikira: I can probably do it for now, as it is only updating the `freeTimeUtility`, but eventually I will need to get the `preAssigned` time variable (of type long) from the `for` loop. Is there a way you know I could do that as well?

src/main/java/com/google/sps/servlets/PlanMailServlet.java
AymarMuhikira: There are basically two ways to get the `DateTime` from an `Event`, either `getDateTime` or `getDate`. What this is doing is making sure we account for both possibilities for each event.

src/main/java/com/google/sps/servlets/PlanMailServlet.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/PlanMailServlet.java
AymarMuhikira: `Resolved`

src/main/webapp/populateData.js
AymarMuhikira: Me neither. But using a `for-in` loop, it is necessary to do it to pass the `lint.js` check

src/test/java/ClientIDServletTest.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/data/PlanMailResponse.java
```diff
@@ -0,0 +1,55 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import com.google.sps.utility.DateInterval;
+import java.util.*;
```
tttzach, line 18: Change this to `import java.util.List;`, the style guide [doesn't like wildcard imports](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports).

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -0,0 +1,209 @@
...
+    long fiveDaysInMillis = TimeUnit.DAYS.toMillis(5);
+    Date timeMin = calendarClient.getCurrentTime();
+    Date timeMax = Date.from(timeMin.toInstant().plus(Duration.ofDays(5)));
+    List<Event> calendarEvents = getEvents(calendarClient, timeMin, timeMax);
+
+    int personalBeginHour = 7;
+    int workBeginHour = 10;
+    int workEndHour = 18;
+    int personalEndHour = 23;
+    int numDays = 5;
```
tttzach, line 102: Maybe a comment to explain these numbers? I can predict what `numDays` means but not so much the others.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -0,0 +1,209 @@
...
+    FreeTimeUtility freeTimeUtility =
+        new FreeTimeUtility(
+            timeMin, personalBeginHour, workBeginHour, workEndHour, personalEndHour, numDays);
+    for (Event event : calendarEvents) {
+      DateTime start = event.getStart().getDateTime();
+      start = start == null ? event.getStart().getDate() : start;
+      DateTime end = event.getEnd().getDateTime();
+      end = end == null ? event.getEnd().getDate() : end;
```
tttzach, line 110: Nice use of ternary operators

src/main/webapp/populateData.js
```diff
@@ -364,6 +364,48 @@ function populateGo() {
...
+            const liElement = document.createElement('li');
+            liElement.innerText =
+              planMailResponse.potentialMeetingTimes[index].start +
+              ' ' +
+              planMailResponse.potentialMeetingTimes[index].end;
```
tttzach, line 396: Oh I remember this, I am guessing the later branch replaces this?

src/main/webapp/auth.js
```diff
@@ -38,7 +38,7 @@
 /* eslint-disable no-unused-vars */
 /* global gapi, addCookie, isCookiePresent,
 deleteCookie, populateGmail, populateTasks, populateCalendar,
-postAndGetTaskList, populateGo, setUpAssign */
+postAndGetTaskList, populateGo, populatePlanMail,setUpAssign */
```
tttzach, line 41: nit: whitespace after ,

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -0,0 +1,209 @@
...
+   * @return List of Events from all of the user's calendars
+   * @throws IOException if an issue occurs in the method
+   */
+  private List<Event> getEvents(CalendarClient calendarClient, Date timeMin, Date timeMax)
+      throws IOException {
+    List<CalendarListEntry> calendarList = calendarClient.getCalendarList();
+    List<Event> events = new ArrayList<>();
+    for (CalendarListEntry calendar : calendarList) {
+      events.addAll(calendarClient.getUpcomingEvents(calendar, timeMin, timeMax));
+    }
```
tttzach, line 187: Would it be possible to do `calendarList.stream().map(calendar -> events.addAll(calendarClient.getUpcomingEvents(calendar, timeMin, timeMax)))`?

src/main/java/com/google/sps/data/PlanMailResponse.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: Thank you

src/main/webapp/populateData.js
AymarMuhikira: Yes, It is fixed in PR #150

src/main/webapp/auth.js
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: Not exactly like that, handling of exception requires to increase the expression when using this method. I think it might get more complicated instead of easier

src/main/webapp/populateData.js
austinteshuba: That's strange - removing the `if (typeof index == 'string')` shouldn't cause that issue. 

If this has to stay, use `===` then to avoid type coercion

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -14,49 +14,67 @@
...
+   * Initialize the class with the start day. The work hours and personal hours are provided as
+   * parameters of the constructor.
    *
    * @param startTime parameter that gives the time of the start/now.
+   * @param personalBeginHour parameter that gives the hour to begin the personal time
+   * @param workBeginHour parameter that gives the hour to begin the work time
+   * @param workEndHour parameter that gives the hour to begin the work time
    */
-  public FreeTimeUtility(Date startDate) {
+  public FreeTimeUtility(
```
austinteshuba: I would consider storing the integers in this utility as `public static final` constants, since they are used in other files as well. That way, `personalBeginHour`, `workBeginHour`, etc. do not need to be defined in each servlet

src/main/java/com/google/sps/utility/FreeTimeUtility.java
```diff
@@ -74,73 +92,107 @@ public void addEvent(Date eventStart, Date eventEnd) {
...
-      } else if (interval.getKey().before(eventStart) && interval.getValue().after(eventStart)) {
-        updatedTime.add(new DatePair(interval.getKey(), eventStart));
-      } else if (interval.getKey().before(eventEnd) && interval.getValue().after(eventEnd)) {
-        updatedTime.add(new DatePair(eventEnd, interval.getValue()));
-      } else if (!(eventStart.before(interval.getKey()) && eventEnd.after(interval.getValue()))) {
+  private List<DateInterval> updateInterval(
+      List<DateInterval> freeTime, Date eventStart, Date eventEnd) {
+    List<DateInterval> updatedTime = new ArrayList<>();
+    for (DateInterval interval : freeTime) {
+      // interval:      |--------------|
```
austinteshuba: I like these ASCII diagrams!

AymarMuhikira: Thank you!

src/main/java/com/google/sps/utility/FreeTimeUtility.java
AymarMuhikira: The integers you are mentioning are parameters to the constructor, but not global variables of the class. The reason they are parameters is so that when scaling the project, the user can have the option to set these values from the front-end, and so the values would be served from the request.

austinteshuba: Ah okay yes that makes sense - they won't be hardcoded forever. 

Thanks!

src/main/java/com/google/sps/data/PlanMailResponse.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.sps.utility.DateInterval;
+import java.util.List;
+
+/** Class containing the response to be converted to Json. */
+public final class PlanMailResponse {
+
+  private final int wordCount;
+  private final int averageReadingSpeed;
+  private final int minutesToRead;
+  private final List<DateInterval> potentialMeetingTimes;
```
antmar, line 26: Nice re-use of `DateInterval` data structure in `PlanMailResponse` !

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
antmar: Thanks for adding the documentation! I think these should be class constants for now - you can always change that if/when you supply some of the values in the request.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -0,0 +1,211 @@
...
+    // work hours as defined between 10am and 6 pm. The rest of the time between 7 am and 11 pm
+    // should be considered personal time.
+    int personalBeginHour = 7;
+    int workBeginHour = 10;
+    int workEndHour = 18;
+    int personalEndHour = 23;
+    int numDays = 5;
+    FreeTimeUtility freeTimeUtility =
+        new FreeTimeUtility(
+            timeMin, personalBeginHour, workBeginHour, workEndHour, personalEndHour, numDays);
```
antmar, line 107: Nice re-use of `FreeTimeUtility` !

src/main/webapp/populateData.js
```diff
@@ -340,3 +340,105 @@ function populateGo() {
...
+        // If response is a 403, user is not authenticated
+        if (response.status === 403) {
+          throw new AuthenticationError();
+        }
+        return response.json();
+      })
+      .then((planMailResponse) => {
+        // Display the potential times to create events
+        const numWordsContainer = document.querySelector('#word-count');
+        numWordsContainer.innerText = planMailResponse.wordCount;
```
antmar, line 360: Suggestion: if you aren't going to use variables like `numWordsContainer` again, then just use the result of the `document.querySelector` inline:
```js
document.querySelector('#word-count').innerText = planMailResponse.wordCount;
```

src/main/webapp/populateData.js
antmar: Did you figure out what's going on here? I also pointed this out in another PR - it seems strange that the linter requires this check. (Also agree with using strict equality if this is needed.)

src/main/webapp/populateData.js
```diff
@@ -340,3 +340,105 @@ function populateGo() {
...
+          signOut();
+        }
+      });
+}
+
+/**
+ * Set up the assign panel. For now, this just prints the response
+ * from the server for /gmail-actionable-emails
+ */
+function setUpAssign() {
```
antmar, line 390: I am confused why these assign-related functions are included here. Is this the result of a merge conflict?

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
antmar: What about using `Stream.flatMap` to flatten the (conceptually) nested list of events in the list of calendars? I'll leave the details as an exercise for you.

Alternatively, you could use the idea that @tttzach suggested, but replace `Stream.map` with `Stream.forEach`.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import javax.mail.MessagingException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** GET function responds JSON string containing potential events to read mail. */
+@WebServlet("/plan-mail")
+public class FreeTimeServlet extends AuthenticatedHttpServlet {
```
antmar, line 52: Can we call this `PlanMailServlet` so it's more consistent with both the URI and the name of the feature? It's not returning generic free-time information - it's returning event slots for reading mail.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -0,0 +1,211 @@
...
+    this.calendarClientFactory = calendarClientFactory;
+    this.gmailClientFactory = gmailClientFactory;
+  }
+
+  /**
+   * Returns string containing the some of the user's free time intervals, where events can be
+   * created
+   *
+   * @param request Http request from the client. Should contain idToken and accessToken
+   * @param response 403 if user is not authenticated, or Json string with the user's events
```
antmar, line 82: Doc comment: technically speaking, this method should never be called is the response contains a 403 error. In this method, you  are always returning a Json string in the response.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -0,0 +1,211 @@
...
+        eventStart = timeMin;
+      }
+      if (eventEnd.after(timeMax)) {
+        eventEnd = timeMax;
+      }
+      freeTimeUtility.addEvent(eventStart, eventEnd);
+    }
+
+    int wordCount = getWordCount(googleCredential);
+    int averageReadingSpeed = 50;
```
antmar, line 125: Also make this a class constant for now.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -0,0 +1,211 @@
...
+        new PlanMailResponse(wordCount, averageReadingSpeed, minutesToRead, potentialTimes);
+    // Convert event list to JSON and print to response
+    JsonUtility.sendJson(response, planMailResponse);
+  }
+
+  private List<DateInterval> getPotentialTimes(FreeTimeUtility freeTimeUtility, long timeNeeded) {
+    List<DateInterval> workFreeInterval = freeTimeUtility.getWorkFreeInterval();
+    List<DateInterval> potentialEventTimes = new ArrayList<>();
+    long remainingTime = timeNeeded;
+    for (DateInterval interval : workFreeInterval) {
```
antmar, line 140: This loop has a relatively complicated exit condition, so you should leave a short comment documenting it. 

It helps to talk in terms of *loop invariants*. For example, `remainingTime` is >0 in the loop, and something about the content of `potentialEventTimes`.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    int wordCount = getWordCount(googleCredential);
+    int averageReadingSpeed = 50;
+    int minutesToRead = (int) Math.ceil((double) wordCount / averageReadingSpeed);
+    long timeNeeded = minutesToRead * TimeUnit.MINUTES.toMillis(1);
+    List<DateInterval> potentialTimes = getPotentialTimes(freeTimeUtility, timeNeeded);
+
+    PlanMailResponse planMailResponse =
+        new PlanMailResponse(wordCount, averageReadingSpeed, minutesToRead, potentialTimes);
```
antmar, line 131: Do you indicate in any way if there is not enough time to read all the e-mails? I just want to make sure I am not missing anything...

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -0,0 +1,211 @@
...
+  private int getWordCount(Credential googleCredential) {
+    GmailClient gmailClient = gmailClientFactory.getGmailClient(googleCredential);
+    GmailClient.MessageFormat messageFormat = GmailClient.MessageFormat.FULL;
+    int numberDays = 7;
+    int wordCount = 0;
+    List<Message> unreadMessages = new ArrayList<>();
+    try {
+      unreadMessages = gmailClient.getUnreadEmailsFromNDays(messageFormat, numberDays);
+    } catch (IOException e) {
+      System.out.println(e);
```
antmar, line 162: You probably want to wrap the exception in some custom type and re-throw it here. Otherwise the user will just think they have no unread e-mails to deal with.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+    for (Message message : unreadMessages) {
+      try {
+        wordCount += getMessageSize(message);
+      } catch (MessagingException | IOException e) {
+        System.out.println(e);
```
antmar, line 168: Similarly, I think you need to re-throw or wrap these exceptions?

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -0,0 +1,211 @@
...
+  /**
+   * Get the events in the user's calendars
+   *
+   * @param calendarClient either a mock CalendarClient or a calendarClient with a valid credential
+   * @param timeMin the minimum time to start looking for events
+   * @param timeMax the maximum time to look for events
+   * @return List of Events from all of the user's calendars
+   * @throws IOException if an issue occurs in the method
+   */
+  private List<Event> getEvents(CalendarClient calendarClient, Date timeMin, Date timeMax)
```
antmar, line 183: Doc comment: you are being inconsistent with documenting (or not documenting) your private methods. I suggest writing documentation for each, because they are not trivial.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -0,0 +1,211 @@
...
+    return events;
+  }
+
+  private int getMessageSize(Message message) throws MessagingException, IOException {
+    List<MessagePart> messageParts = message.getPayload().getParts();
+    List<MessagePart> messageBody =
+        messageParts.stream()
+            .filter((messagePart) -> messagePart.getMimeType().equals("text/plain"))
+            .collect(Collectors.toList());
+    int size = 0;
```
antmar, line 199: I think you could do everything in this method with a single stream operation. Take a look at `Stream.flatMap` and `Stream.reduce` for inspiration.

antmar:
Ah, I see you already moved the helpers out of the servlet in PR #154. That was going to be my other comment :-)

src/main/java/com/google/sps/data/PlanMailResponse.java
AymarMuhikira: Thank you.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: Thank you!

src/main/webapp/populateData.js
AymarMuhikira: `Resolved`

src/main/webapp/populateData.js
AymarMuhikira: Yes. I did not implement this. They appeared when updating the branch with master

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
antmar: I just noticed that you originally called this PlanMailServlet, and @austinteshuba suggested you change this. Apologies for the conflicting feedback. 

However, I stand by my claim that it's not serving generic free-time information, but rather specific mail-reading-related event slots. (This is even more true once the creation of calendar events is implemented, and this actually looks for/ignores existing events for reading mail.)

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: The servlet proposes all possible event times until it is not possible. For now, the user does not get an indication if the events proposed are sufficient to go through the emails. I will add a comment on this

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: I added a Javadoc description of the method

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: I used `Throwables` to re-throw the exceptions.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: `Resolved`

PR #131, opened 2020-07-31
Implement Go servlet

tttzach:
Initial implementation of the Go servlet.


src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,301 @@
...
+        streetAddressWaypointsAsCoordinates,
+        nonStreetAddressWaypointsAsPlaceTypes);
+
+    // All street address coordinates including origin and destination are collected
+    List<LatLng> streetAddressesAsCoordinates = streetAddressWaypointsAsCoordinates;
+    streetAddressesAsCoordinates.add(originAsCoordinates);
+    streetAddressesAsCoordinates.add(destinationAsCoordinates);
+
+    // Remove all null entries of street address coordinates and non street address waypoints
+    streetAddressesAsCoordinates =
```
austinteshuba, line 279: Why would there be null entries here? I would add a comment explaining what the represent, and possibly try to eliminate them earlier on (i.e. if they are being added from the `seperateWaypoints` method, I would advise adding a null check in that method to avoid this issue, and if this is a result of `originAsCoordinates` or `destinationAsCoordinates` being null, I would avoid adding them to begin with if they are null and possibly return early if this results in failure)

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,301 @@
...
+   */
+  public List<List<String>> searchNearbyEveryKnownLocationForClosestPlaceTypeMatch(
+      List<PlaceType> nonStreetAddressWaypointsAsPlaceTypes,
+      List<LatLng> streetAddressesAsCoordinates)
+      throws PlacesException {
+    List<List<String>> allSearchNearbyResults = new ArrayList<>();
+    for (PlaceType nonStreetAddressWaypoint : nonStreetAddressWaypointsAsPlaceTypes) {
+      List<String> searchNearbyResults = new ArrayList<>();
+      for (LatLng coordinate : streetAddressesAsCoordinates) {
+        PlaceType placeType = nonStreetAddressWaypoint;
```
austinteshuba, line 196: No need for the variable `placeType`

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,301 @@
...
+      String origin, String destination, List<List<String>> allWaypointCombinations)
+      throws DirectionsException {
+    long minTravelTime = 0;
+    List<String> mostOptimalWaypointCombination = new ArrayList<String>();
+    for (List<String> waypointCombination : allWaypointCombinations) {
+      DirectionsClient directionsClient = directionsClientFactory.getDirectionsClient(apiKey);
+      DirectionsResult directionsResult =
+          directionsClient.getDirections(origin, destination, waypointCombination);
+      long travelTime = DirectionsClient.getTotalTravelTime(directionsResult);
+      if (minTravelTime == 0 || travelTime < minTravelTime) {
```
austinteshuba, line 230: Use `Long.MAX_VALUE` for the initial value of `minTravelTime`, and then you can simplify this check to be `Math.min(minTravelTime, travelTime)`

src/test/java/GoServletTest.java
```diff
@@ -0,0 +1,254 @@
...
+  private DirectionsRoute shorterRoute = new DirectionsRoute();
+  private DirectionsResult shorterResult = new DirectionsResult();
+
+  @Override
+  @Before
+  public void setUp() throws Exception {
+    super.setUp();
+    directionsClientFactory = Mockito.mock(DirectionsClientFactory.class);
+    directionsClient = Mockito.mock(DirectionsClient.class);
+    placesClientFactory = Mockito.mock(PlacesClientFactory.class);
```
austinteshuba, line 84: You shouldn't have to mock a factory

src/test/java/GoServletTest.java
```diff
@@ -0,0 +1,254 @@
...
+  private DirectionsClient directionsClient;
+  private PlacesClientFactory placesClientFactory;
+  private PlacesClient placesClient;
+  private TasksClientFactory tasksClientFactory;
+  private TasksClient tasksClient;
+  private GeocodingClientFactory geocodingClientFactory;
+  private GeocodingClient geocodingClient;
+  private GoServlet servlet;
+
+  private static final String API_KEY = "fake api key";
```
austinteshuba, line 57: nit: Keep `private` and `private static` variables together (i.e. move the `private` declarations up)

src/test/java/GoServletTest.java
```diff
@@ -0,0 +1,254 @@
...
+    tasksClient = Mockito.mock(TasksClient.class);
+    geocodingClientFactory = Mockito.mock(GeocodingClientFactory.class);
+    geocodingClient = Mockito.mock(GeocodingClient.class);
+
+    streetAddressGeocodingResult.geometry = streetAddressGeometry;
+    streetAddressGeocodingResult.geometry.location = streetAddressCoordinates;
+    streetAddressGeocodingResult.types = new AddressType[] {AddressType.STREET_ADDRESS};
+    restaurantGeocodingResult.types = new AddressType[] {AddressType.RESTAURANT};
+
+    longerDuration.inSeconds = 10;
```
austinteshuba, line 96: Maybe set the time value to be a constant

src/test/java/GoServletTest.java
```diff
@@ -0,0 +1,254 @@
...
+  private static final String RESTAURANT_TWO = "restaurant near coordinate two";
+  private static final String BANK_ONE = "bank near coordinate one";
+  private static final String BANK_TWO = "bank near coordinate two";
+
+  private GeocodingResult streetAddressGeocodingResult = new GeocodingResult();
+  private Geometry streetAddressGeometry = new Geometry();
+  private LatLng streetAddressCoordinates = new LatLng(0, 0);
+  private GeocodingResult restaurantGeocodingResult = new GeocodingResult();
+
+  private Duration longerDuration = new Duration();
```
austinteshuba, line 68: Any reason why these are not static? Seems like they are initialized with the same values each time

src/test/java/GoServletTest.java
```diff
@@ -0,0 +1,254 @@
...
+            directionsClientFactory,
+            placesClientFactory,
+            tasksClientFactory,
+            geocodingClientFactory,
+            API_KEY);
+
+    Mockito.when(directionsClientFactory.getDirectionsClient(API_KEY)).thenReturn(directionsClient);
+    Mockito.when(placesClientFactory.getPlacesClient(API_KEY)).thenReturn(placesClient);
+    Mockito.when(tasksClientFactory.getTasksClient(Mockito.any())).thenReturn(tasksClient);
+    Mockito.when(authenticationVerifier.verifyUserToken(Mockito.anyString()))
```
austinteshuba, line 117: Shouldn't have to specify the mock for `authenticationVerfier` - this should be handled by the super class

src/test/java/GoServletTest.java
```diff
@@ -0,0 +1,254 @@
...
+        .thenReturn(AUTHENTICATION_VERIFIED);
+    Mockito.when(geocodingClientFactory.getGeocodingClient(API_KEY)).thenReturn(geocodingClient);
+  }
+
+  @Test
+  public void separateWaypoints() throws Exception {
+    // Street address should be converted into coordinates, restaurant should be converted into
+    // place type
+    List<String> waypoints = ImmutableList.of("street address", "restaurant");
```
austinteshuba, line 126: Make this a constant, since `streetAddressGeocodingResult` is already a constant. This list, as well as its contents, have semantic value

src/test/java/GoServletTest.java
```diff
@@ -0,0 +1,254 @@
...
+            ImmutableList.of("place_id:" + BANK_ONE, "place_id:" + BANK_TWO)),
+        actual);
+  }
+
+  @Test
+  public void chooseWaypointCombinationWithShortestTravelTime() throws Exception {
+    // Find the waypoints which result in the shorter travel time
+    String origin = "A";
+    String destination = "B";
+    List<String> longerTravelTimeWaypoints = ImmutableList.of("long", "long");
```
austinteshuba, line 184: Ditto on making this a constant

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,301 @@
...
+   * @throws ServletException
+   * @throws IOException
+   */
+  @Override
+  public void doGet(
+      HttpServletRequest request, HttpServletResponse response, Credential googleCredential)
+      throws ServletException, IOException {
+    assert googleCredential != null
+        : "Null credentials (i.e. unauthenticated requests) should already be handled";
+    // Get all tasks from user's tasks account TODO: Get relevant tasks using task
```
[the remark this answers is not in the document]
tttzach: Added PR number since it's done already, thanks for pointing it out!

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,301 @@
...
+    // titles
+    TasksClient tasksClient = tasksClientFactory.getTasksClient(googleCredential);
+    DirectionsClient directionsClient = directionsClientFactory.getDirectionsClient(apiKey);
+
+    // Initialize Tasks Response
+    List<Task> tasks;
+
+    String taskLists = request.getParameter("taskLists");
+    if (taskLists == null) {
+      tasks = TasksServlet.getAllTasksFromAllTaskLists(tasksClient);
```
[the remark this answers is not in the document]
tttzach: Initially yes, but not anymore because these methods have been moved to a utility class in https://github.com/googleinterns/blueprint-step-2020/pull/146.

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,301 @@
...
+    this.placesClientFactory = placesClientFactory;
+    this.tasksClientFactory = tasksClientFactory;
+    this.geocodingClientFactory = geocodingClientFactory;
+    this.apiKey = apiKey;
+  }
+
+  /**
+   * Returns the most optimal order of travel between addresses.
+   *
+   * @param request HTTP request from the client.
```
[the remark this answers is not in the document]
tttzach: Oh good point, did not register query strings as parameters in my head. Thanks!

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,301 @@
...
+    } else {
+      List<String> selectedTaskListIds = Arrays.asList(taskLists.split(","));
+      tasks = TasksServlet.getAllTasksFromSpecificTaskLists(tasksClient, selectedTaskListIds);
+    }
+
+    String origin = request.getParameter("origin");
+    String destination = request.getParameter("destination");
+    List<String> waypoints = LocationsUtility.getLocations("Location", tasks);
+
+    try {
```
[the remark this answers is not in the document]
tttzach: Thanks!

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,301 @@
...
+   *     non street addresses.
+   * @param streetAddressWaypoints A pointer to the result for street address waypoints to achieve
+   *     the effect of returning multiple types at once.
+   * @param streetAddressWaypointsAsCoordinates A pointer to the result for street address waypoints
+   *     as coordinates to achieve the effect of returning multiple types at once.
+   * @param nonStreetAddressWaypointsAsPlaceTypes A pointer to the result for non street address
+   *     waypoints as place types to achieve the effect of returning multiple types at once.
+   * @throws GeocodingException An exception thrown when an error occurs with the Geocoding API.
+   */
+  public void separateWaypoints(
```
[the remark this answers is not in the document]
tttzach: Yeah I agree a lot is going on in this method and I would also like to know if @gbuenoandrade or @antmar has any ideas on how to break this (and the other large methods up). Unfortunately, it's not the looping that's expensive, it's the call to the Geocoding API at line 169.

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,301 @@
...
+    List<LatLng> streetAddressWaypointsAsCoordinates = new ArrayList<>();
+    List<PlaceType> nonStreetAddressWaypointsAsPlaceTypes = new ArrayList<>();
+    separateWaypoints(
+        waypoints,
+        streetAddressWaypoints,
+        streetAddressWaypointsAsCoordinates,
+        nonStreetAddressWaypointsAsPlaceTypes);
+
+    // All street address coordinates including origin and destination are collected
+    List<LatLng> streetAddressesAsCoordinates = streetAddressWaypointsAsCoordinates;
```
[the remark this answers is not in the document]
tttzach: Hmm it would be misleading to add the origin and destination into a list that is meant for waypoints. I've sort of took that direct assignment out and used `addAll` instead. I don't know if that helps, let me know!

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: The null entries are from the other PRs where you suggested empty optionals for. I've added an optional empty check for `originAsCoordinates` and `destinationAsCoordinates` to throw an error if that happens. Thanks for pointing it out!

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: Agreed, thanks!

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: Hmm not sure if I am understanding correctly, I kinda need to reassign `mostOptimalWaypointCombination` so I don't think that would help.

src/test/java/GoServletTest.java
tttzach: Oh true, thanks for pointing that out!

tttzach: Not sure if I am understanding correctly again, kinda need it to give explicit implementations to the GoServlet constructor. The default constructor gets API keys using file I/O.

src/test/java/GoServletTest.java
tttzach: Good idea, thanks!

src/test/java/GoServletTest.java
tttzach: Oh yeah you're right, thanks!

src/test/java/GoServletTest.java
tttzach: Good call, thanks!

src/test/java/GoServletTest.java
tttzach: Uh nope, just forgot - thanks!

src/test/java/GoServletTest.java
tttzach: Good call, thanks!

src/test/java/GoServletTest.java
tttzach: Yep it's used twice too, thanks!

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: `RESOLVED`

src/test/java/GoServletTest.java
tttzach: `RESOLVED`

src/test/java/GoServletTest.java
tttzach: `RESOLVED`

src/test/java/GoServletTest.java
tttzach: `RESOLVED`

src/test/java/GoServletTest.java
tttzach: `RESOLVED`

src/test/java/GoServletTest.java
tttzach: `RESOLVED`

src/test/java/GoServletTest.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
gbuenoandrade: It would remove the need to check if `mintravelTime == 0`. In competitive programming, we always do that. As a matter of fact, there's a magic number we use to represent infinity "0x3f3f3f3f". It is "magic" because it is large, doubling it still fits into a 32 bit integer, and because the bytes of it are the same, you can fill an array with that value using the C function `memset`. Though, I'm not sure if that's considered a "good" practice :( Maybe using `Optional` would be preferred.

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,316 @@
...
+    streetAddressesAsCoordinates.add(originAsCoordinates);
+    streetAddressesAsCoordinates.add(destinationAsCoordinates);
+
+    // Remove all optional empty entries of street address coordinates and non street address
+    // waypoints
+    List<LatLng> nonEmptyStreetAddressesAsCoordinates =
+        streetAddressesAsCoordinates.stream()
+            .filter(coordinates -> !coordinates.equals(Optional.empty()))
+            .map(coordinates -> coordinates.get())
+            .collect(Collectors.toList());
```
gbuenoandrade, line 295: maybe create an auxiliary method to do that? It could accept `List<Optional<T>>` so you can use it for both lists

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,316 @@
...
+      Set<String> selectedTaskListIds = new HashSet<>(Arrays.asList(taskLists.split(",")));
+      tasks = TasksUtility.getAllTasksFromSpecificTaskLists(tasksClient, selectedTaskListIds);
+    }
+
+    String origin = request.getParameter("origin");
+    String destination = request.getParameter("destination");
+    List<String> waypoints = LocationsUtility.getLocations("Location", tasks);
+
+    try {
+      List<String> mostOptimalWaypointCombination =
```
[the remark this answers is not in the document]
tttzach: Agreed, thanks!

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
austinteshuba: Yes, I think this makes it more readable. Thanks!

src/main/java/com/google/sps/servlets/GoServlet.java
austinteshuba: @gbuenoandrade Sounds like a cool number :) +1 on using optional if `Long.MAX_VALUE` is not recommended

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
tttzach: Oh this is a great idea, thanks!

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,316 @@
...
+            GeocodingResultUtility.convertToPlaceType(waypoint));
+      }
+    }
+  }
+
+  /**
+   * Search nearby every street address with known coordinates for a place type match. For example,
+   * if we know exactly where our houses are and we are looking for a restaurant, we search for a
+   * restaurant closest to your house and a restaurant closest to my house. In this case, the method
+   * should return [[restaurantOne, restaurantTwo]]. Scope of method is public for testing purposes.
```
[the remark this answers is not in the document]
tttzach: Yeah I agree with you, I could have definitely moved that into a class on its own and abstracted out the API call. In the interest of time, I will just imprint that into my mind for the next internship. :(

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,316 @@
...
+  public List<List<String>> searchNearbyEveryKnownLocationForClosestPlaceTypeMatch(
+      List<PlaceType> nonStreetAddressWaypointsAsPlaceTypes,
+      List<LatLng> streetAddressesAsCoordinates)
+      throws PlacesException {
+    List<List<String>> allSearchNearbyResults = new ArrayList<>();
+    for (PlaceType nonStreetAddressWaypoint : nonStreetAddressWaypointsAsPlaceTypes) {
+      List<String> searchNearbyResults = new ArrayList<>();
+      for (LatLng coordinate : streetAddressesAsCoordinates) {
+
+        PlacesClient placesClient = placesClientFactory.getPlacesClient(apiKey);
```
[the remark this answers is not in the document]
tttzach: Sadly, it doesn't work because it needs a new client each time otherwise it throws `java.lang.IllegalStateException: 'await', 'awaitIgnoreError' or 'setCallback' was already called.`. Same with the Geocoding client. :/

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,316 @@
...
+   *     restaurant, supermarket, police station)
+   * @param streetAddressesAsCoordinates A list of coordinates to look for the place types around.
+   * @return A list of lists of place IDs where each list represents the search nearby result for
+   *     every known coordinate.
+   * @throws PlacesException An exception thrown when an error occurs with the Places API.
+   */
+  public List<List<String>> searchNearbyEveryKnownLocationForClosestPlaceTypeMatch(
+      List<PlaceType> nonStreetAddressWaypointsAsPlaceTypes,
+      List<LatLng> streetAddressesAsCoordinates)
+      throws PlacesException {
```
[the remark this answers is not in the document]
tttzach: True, thanks! That method name is a lot shorter.

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
tttzach: Awesome, thanks!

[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: Optional is a great idea, I think it gives more information. Thanks!

tttzach: minTravelTime.isPresent()

tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: Gonna ask you on chat - not entirely sure how to set types in Java templates

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
gbuenoandrade: Gotcha
RESOLVED

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,332 @@
...
+    // All street address coordinates including origin and destination are collected
+    List<Optional<LatLng>> streetAddressesAsCoordinates = new ArrayList<>();
+    streetAddressesAsCoordinates.addAll(separatedWaypoints.streetAddressWaypointsAsCoordinates);
+    streetAddressesAsCoordinates.add(originAsCoordinates);
+    streetAddressesAsCoordinates.add(destinationAsCoordinates);
+
+    // Remove all optional empty entries of street address coordinates and non street address
+    // waypoints
+    List<LatLng> nonEmptyStreetAddressesAsCoordinates = filterNonNull(streetAddressesAsCoordinates);
+    List<PlaceType> nonEmptynonStreetAddressWaypointsAsPlaceTypes =
```
gbuenoandrade, line 315: nit: nonEmptyNonStreet

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -0,0 +1,332 @@
...
+        searchForPlacesNearLocations(
+            nonEmptynonStreetAddressWaypointsAsPlaceTypes, nonEmptyStreetAddressesAsCoordinates);
+
+    List<List<String>> allWaypointCombinations =
+        LocationsUtility.generateCombinations(allSearchNearbyResults);
+
+    List<String> optimalWaypointCombination =
+        chooseWaypointCombinationWithShortestTravelTime(
+            origin, destination, allWaypointCombinations);
+    optimalWaypointCombination.addAll(separatedWaypoints.streetAddressWaypoints);
```
gbuenoandrade, line 328: why is this necessary?

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
gbuenoandrade: nit: I think you could also rename the parameters and local variables, but it's up to you

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: Uh oh, thanks!

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: Oh not sure why I left that outside the function. Thanks for catching that!

tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: Fixed, thanks! Like you mentioned, was just missing a <T>

tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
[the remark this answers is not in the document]
tttzach: I renamed them, I agree they don't have to be as long in this method. Thanks!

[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -209,21 +210,20 @@ public GoResponse separateWaypoints(List<String> waypoints) throws GeocodingExce
...
    * @throws PlacesException An exception thrown when an error occurs with the Places API.
    */
   public List<List<String>> searchForPlacesNearLocations(
-      List<PlaceType> nonStreetAddressWaypointsAsPlaceTypes,
-      List<LatLng> streetAddressesAsCoordinates)
+      List<PlaceType> placeTypeWaypoints, List<LatLng> streetAddressWaypoints)
       throws PlacesException {
     List<List<String>> allSearchNearbyResults = new ArrayList<>();
-    for (PlaceType nonStreetAddressWaypoint : nonStreetAddressWaypointsAsPlaceTypes) {
+    for (PlaceType nonStreetAddressWaypoint : placeTypeWaypoints) {
```
gbuenoandrade, line 224: Super nit: the for variable now looks a bit confusing

src/main/java/com/google/sps/utility/GeocodingResultUtility.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Optional;
+
+/** Utility class to extract data from GeocodingResult objects. */
+public class GeocodingResultUtility {
+  /**
+   * Parses for the first coordinate which is of a street address type in a result from the
+   * Geocoding API. If no street address type results are found, the coordinates of the first result
+   * is returned.
+   *
+   * @param result A GeocodingResult returned from the Geocoding API.
```
austinteshuba, line 32: Javadoc does not match the header here - change parameter name

tttzach: Done, thanks!

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: Agreed, thanks!

tttzach: `RESOLVED`

src/main/java/com/google/sps/utility/GeocodingResultUtility.java

PR #134, opened 2020-07-31
Remove double authentication calls

austinteshuba:
Also added some console logs to mark when the system is signing in, singing out, and populating the panels. 

Let me know your thoughts on keeping the logs - I did want to include them in this PR (so you can play around with auth and check that this solution works), and I wouldn't mind keeping them for the final version (they're written with good language and shouldn't be an issue to the user), but I also understand if we want to remove them.

Solves issue #97


austinteshuba:
> Can you document somewhere how the auth for a signed-in user works? It's not trivial to follow.
> 
> Here is my understanding:
> 
> * `init` fetches the client ID, initializes `gapi.auth2`, and calls `renderButton`
> * `renderButton` renders the button, which may be in a signed-in or signed-out state
>   
>   * for a signed-in state, it calls `onSignIn` automatically
> * `onSignIn` eventually makes the call to load the data

Just updated the README - let me know if this helps/is appropriate

antmar:
> > Can you document somewhere how the auth for a signed-in user works? It's not trivial to follow.
> > Here is my understanding:
> > 
> > * `init` fetches the client ID, initializes `gapi.auth2`, and calls `renderButton`
> > * `renderButton` renders the button, which may be in a signed-in or signed-out state
> >   
> >   * for a signed-in state, it calls `onSignIn` automatically
> > * `onSignIn` eventually makes the call to load the data
> 
> Just updated the README - let me know if this helps/is appropriate

Summarizing discussion in chat - moved this to a code comment in `auth.js` because this is an implementation detail. 

Thanks for adding the documentation!

PR #136, opened 2020-07-31
Implement Geocoding client

tttzach:
Implement Geocoding client to determine get coordinates and place types of addresses for the Go panel.


src/main/java/com/google/sps/model/GeocodingClient.java
```diff
@@ -0,0 +1,87 @@
...
+   */
+  GeocodingResult getGeocodingResult(String address) throws GeocodingException;
+
+  /**
+   * Parses for coordinates in a resulting call to the Geocoding API.
+   *
+   * @param result A GeocodingResult returned from the Geocoding API.
+   * @return A LatLng representing coordinates.
+   */
+  public static LatLng getCoordinates(GeocodingResult result) {
```
austinteshuba, line 44: nit: `public` is redundant in interfaces

src/main/java/com/google/sps/model/GeocodingClient.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  /**
+   * Converts an AddressType to a PlaceType for the purpose of calling the Places API. PlaceType is
+   * a subset of AddressType and hence, not all AddressTypes are supported by the Places API.
+   *
+   * @param addressType An AddressType to convert to a PlaceType
+   * @return A PlaceType corresponding to an AddressType if available, null if not available.
+   */
+  public static PlaceType convertAddressTypeToPlaceType(AddressType addressType) {
```
austinteshuba, line 55: nit: Use an Optional here if null is a possibility

src/main/java/com/google/sps/model/GeocodingClient.java
tttzach: I keep forgetting this - thanks!

src/main/java/com/google/sps/model/GeocodingClient.java
tttzach: I don't know if that's the best use for [Optional](https://medium.com/faun/working-on-null-elegantly-with-java-optional-62f5e65869c5#:~:text=Optional%20is%20primarily%20intended%20for,point%20to%20an%20Optional%20instance.). I could see how it is useful if you're going to access the returned object to avoid a NullPointerException but this method returns a PlaceType enum and nulls are filtered shortly after in the Go servlet where this is used. I've also added a sentence to the javadoc to somewhat reflect this if that helps.

antmar: I think this would be a great use of `Optional`, because you expect to have address types that don't map to a place type, and there is no natural way to represent "not a valid type" using `PlaceType` itself.

(Sometimes enums will have some sort of UNKNOWN value you could use for this, but I don't think there is such a place type.)

src/main/java/com/google/sps/model/GeocodingClient.java
antmar: All these static methods should live in a `GeocodingResultUtil` or similarly named class. Then the tests would live in a more appropriately-named test class. (These methods have nothing to do with the client, and you don't actually have any tests for the implementation of the client.)

src/main/java/com/google/sps/model/GeocodingClientImpl.java
```diff
@@ -0,0 +1,52 @@
...
+      GeocodingApiRequest service = GeocodingApi.newRequest(context);
+      return new GeocodingClientImpl(service);
+    }
+  }
+
+  @Override
+  public GeocodingResult getGeocodingResult(String address) throws GeocodingException {
+    try {
+      GeocodingResult[] response = geocodingService.address(address).await();
+      return response[0];
```
antmar, line 47: Why would you have multiple results? What does it mean to have no results at all? This should be documented in the method documentation in `GeocodingClient` interface.

Would it make more sense to return the entire list to the servlet, and let it decide what these situations mean?

src/main/java/com/google/sps/model/GeocodingClient.java
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Converts an AddressType to a PlaceType for the purpose of calling the Places API. PlaceType is
+   * a subset of AddressType and hence, not all AddressTypes are supported by the Places API.
+   *
+   * @param addressType An AddressType to convert to a PlaceType
+   * @return A PlaceType corresponding to an AddressType if available, null if not available.
+   */
+  static PlaceType convertAddressTypeToPlaceType(AddressType addressType) {
+    for (PlaceType placeType : PlaceType.values()) {
+      if (placeType.toString().equals(addressType.toString())) {
```
gbuenoandrade, line 57: nit: `.name()` instead of `toString()`

src/main/java/com/google/sps/model/GeocodingClient.java
```diff
@@ -0,0 +1,87 @@
...
+        return placeType;
+      }
+    }
+    return null;
```
gbuenoandrade, line 61: +1 to using `Optional`

src/main/java/com/google/sps/model/GeocodingClient.java
tttzach: Oh okay, so it would be used in place of `PlaceType.UNKNOWN` instead of what the article I linked suggested to use Optional for i.e. avoiding null check validation. I think that makes sense, thanks!

(There's one for AddressType but unfortunately not for PlaceType)

src/main/java/com/google/sps/model/GeocodingClient.java
tttzach: Agreed, I should probably do the same for a lot of the PRs I have since I put most of the methods into the clients as long as they're remotely related (i.e. uses the same API). Test names would make a lot more sense as well since before the changes they're all in GeocodingClientImplTest - which makes no sense.

src/main/java/com/google/sps/model/GeocodingClientImpl.java
tttzach: The Geocoding API could return multiple results in case of partial matches and in that case the first match is the best match. No results would still return a GeocodingResult array (should probably handle this just in case regardless of the assumptions). This is under the assumption that origin and destination provided by the user are valid addresses (the only inputs that do not go through other sorts of validation) because the Maps API does not handle address validation. I agree that this should be documented as well but I don't think there's going to be any use of the rest of the results in the event of partial matches. Let me know what your thoughts are!

src/main/java/com/google/sps/model/GeocodingClient.java
tttzach: Good point, thanks!

src/main/java/com/google/sps/model/GeocodingClient.java
tttzach: Thanks!

src/main/java/com/google/sps/utility/GeocodingResultUtility.java
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Parses for a address type and converts it to a place type if available. This place type is used
+   * to further determine a location which results in a route with the shortest travel time.
+   *
+   * @param result A GeocodingResult returned from the Geocoding API.
+   * @return A PlaceType representing the type or null if no corresponding PlaceType is found.
+   */
+  public static Optional<PlaceType> getPlaceType(GeocodingResult result) {
+    return convertAddressTypeToPlaceType(result.types[0]);
```
gbuenoandrade, line 72: Can `result.types.length` ever be 0? If so, would it make sense to return `Optional.empty()`?

src/main/java/com/google/sps/model/GeocodingClient.java
```diff
@@ -26,62 +23,15 @@
...
+   * Geocoding API could return multiple results in case of partial matches and in that case, the
+   * first match is the best match hence it would be the result returned. No results would cause a
+   * GeocodingException to be thrown.
    *
    * @param address A String representing the address to geocode.
    * @return A GeocodingResult returned from the Geocoding API.
    * @throws GeocodingException A custom exception is thrown to signal an error pertaining to the
    *     Geocoding API.
    */
   GeocodingResult getGeocodingResult(String address) throws GeocodingException;
```
antmar, line 36: Consider changing the name to indicate it returns the best match of all possible results. Something like `getBestGeocodingResult`.

src/main/java/com/google/sps/utility/GeocodingResultUtility.java
```diff
@@ -0,0 +1,74 @@
...
+   * @param addressType An AddressType to convert to a PlaceType
+   * @return A PlaceType corresponding to an AddressType if available, null if not available.
+   */
+  public static Optional<PlaceType> convertAddressTypeToPlaceType(AddressType addressType) {
+    for (PlaceType placeType : PlaceType.values()) {
+      if (placeType.name().equals(addressType.name())) {
+        return Optional.ofNullable(placeType);
+      }
+    }
+    return Optional.empty();
```
antmar, line 48: There is a more succinct way to write this using streams:

```java
placeType.values().stream()
  .filter(t -> t.name().equals(addressType.name()))
  .findFirst();
```
In particular, the stream method `findFirst` returns an Optional.empty() if the stream is empty.

There is also `findAny`, which is similar to `findFirst`, and may be more appropriate here semantically. The place type values aren't really ordered in a way that you care about.

Caveat: I am not sure if `placeType.values()` is a Collection or an Array, and whether or not Arrays support stream.

src/main/java/com/google/sps/model/GeocodingClientImpl.java
antmar: I agree regarding explicitly checking the length of `response` and returning a more helpful error message, like actually saying there were no matches.

src/main/java/com/google/sps/utility/GeocodingResultUtility.java
```diff
@@ -0,0 +1,74 @@
...
+  public static boolean isPartialMatch(GeocodingResult result) {
+    return result.partialMatch;
+  }
+
+  /**
+   * Parses for a address type and converts it to a place type if available. This place type is used
+   * to further determine a location which results in a route with the shortest travel time.
+   *
+   * @param result A GeocodingResult returned from the Geocoding API.
+   * @return A PlaceType representing the type or null if no corresponding PlaceType is found.
```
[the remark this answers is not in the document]
tttzach: Uh oh, thanks!

src/main/java/com/google/sps/utility/GeocodingResultUtility.java
tttzach: I'll make sure to do this in https://github.com/googleinterns/blueprint-step-2020/pull/144 before pushing because this method here is replaced in that PR. Thanks!

src/main/java/com/google/sps/utility/GeocodingResultUtility.java
tttzach: Oh wow this is nice, I've tried it and I just needed to cast `PlaceType[]` to `List<PlaceType>`. Thanks!

src/main/java/com/google/sps/model/GeocodingClientImpl.java
tttzach: I agree, I'll be sure to throw an error message on any index related access to `response` in https://github.com/googleinterns/blueprint-step-2020/pull/144 before merging because this is part here is the only one that accesses response in the PR and it is replaced to return a list in https://github.com/googleinterns/blueprint-step-2020/pull/144. I would like to keep the list because some of the methods iterate through the list of results to find the "best" one.

src/main/java/com/google/sps/model/GeocodingClient.java
tttzach: I like this idea a lot, however, it's going to be replaced in the [next PR](https://github.com/googleinterns/blueprint-step-2020/pull/144) with a method that returns `List<GeocodingResult>` which is useful for some of the methods in https://github.com/googleinterns/blueprint-step-2020/pull/144 so unfortunately it wouldn't apply anymore. Thanks though!

src/main/java/com/google/sps/utility/GeocodingResultUtility.java

src/main/java/com/google/sps/model/GeocodingClient.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/utility/GeocodingResultUtility.java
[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/GeocodingClient.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/GeocodingClient.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/GeocodingClientImpl.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/GeocodingClient.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/GeocodingClient.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/utility/GeocodingResultUtility.java
tttzach: `RESOLVED` https://github.com/googleinterns/blueprint-step-2020/pull/144/commits/ef1d36c9335e2a65269206a97428b3aebd1be44f

PR #132, opened 2020-07-31
Refactor TasksServlet and TaskListsServlet

tttzach:
Abstracts out Tasks panel logic into a new servlet - TasksSummaryServlet. Replaces doGet method of TasksServlet to get all tasks and send them as json. Refactor TasksListsServlet to have a similar structure to TasksServlet.

PR #137, opened 2020-08-01
Working version of the assign panel

austinteshuba:
Finished version of Assign Panel (prioritization on server side still to come). Screenshots below!

Adding new task (hovering over add button): 
Screenshot 2020-08-01 at 8 01 39 PM

Gmail / Tasks view: 
Screenshot 2020-08-01 at 8 04 15 PM

Settings View: 
Screenshot 2020-08-01 at 8 01 56 PM

Dependent on PR #153


austinteshuba:
@austinteshuba do not merge until PR #132 is integrated in this (can only be done once PR #132 is merged to master)

src/main/webapp/task.js
```diff
@@ -25,12 +25,12 @@ class Task {
...
    * @param {string} title a title for the task
    * @param {string} notes messages associated with a given task
-   * @param {Date} due a Date object / null if not present
+   * @param {Date} due a Date object / undefined if not present
    */
   constructor(title, notes, due) {
     this.title = title;
     this.notes = notes;
-    if (due !== null) {
+    if (due != null) {
```
tttzach, line 33: `!==` for consistency

src/main/webapp/assignPanel.js
```diff
@@ -0,0 +1,272 @@
...
+ *     case of other error code
+ */
+function fetchActionableEmails(listOfPhrases, unreadOnly, nDays) {
+  const listOfPhrasesString = encodeListForUrl(listOfPhrases);
+  const unreadOnlyString = unreadOnly.toString();
+  const nDaysString = nDays.toString();
+
+  const queryString =
+      `/gmail-actionable-emails?subjectLinePhrases=${listOfPhrasesString}` +
+      `&unreadOnly=${unreadOnlyString}&nDays=${nDaysString}`;
```
tttzach, line 220: `+` is not needed here because you are using template literals

src/main/webapp/assignPanel.js
```diff
@@ -0,0 +1,272 @@
...
+
+const taskNotesPrefix =
+    '*** ' +
+    'Made automatically with Blueprint. ' +
+    'Please add notes above this line, ' +
+    'and do not edit anything below this line. ' +
+    'Blueprint needs this information to better serve you. ' +
+    'Thank you for your understanding. ' +
+    '***' +
+    '\n';
```
tttzach, line 39: Use template literals here

austinteshuba:
@austinteshuba Refresh tasks once a new task is added

src/main/webapp/task.js
austinteshuba: `!==` and `!=` are actually different operators, where `!=` allows for implicit type coercion and `!==` does not. Fun fact: this is why `true == 1` returns `true` but `true === 1` returns false. 

Usually `!=` is frowned upon, except for catching both `null` and `undefined` values, which is what I'm using this for here. https://google.github.io/styleguide/jsguide.html#features-equality-checks-exceptions

src/main/webapp/assignPanel.js
austinteshuba: I did this intentionally as I did not want a newline character in the string (these are added automatically in a multi-line string made with the backticks)

src/main/webapp/assignPanel.js
austinteshuba: Same reason as above for not using template literals

src/main/webapp/assignPanel.js
```diff
@@ -0,0 +1,272 @@
...
+  while
+  (emailIdValue.substr(0, 'emailId:'.length) !== 'emailId:' && index > 0) {
+    emailIdValue = notes.charAt(index-1) + emailIdValue;
+    index--;
+  }
+  if (index < 0) {
+    console.log('No email ID found!');
+    return null;
+  }
+  return emailIdValue.split(':')[1];
```
gbuenoandrade, line 104: Would [lastIndexOf()](https://www.w3schools.com/jsref/jsref_lastindexof.asp) help here? (or just `indexOf()`, `search()`)

src/main/webapp/assignPanel.js
```diff
@@ -0,0 +1,272 @@
...
+                .filter((taskList) => taskList.title === assignTaskListTitle);
+        if (assignTaskLists.length === 0) {
+          return postNewTaskList(assignTaskListTitle).
+              then((taskList) => {
+                return {
+                  'taskListId': taskList.id,
+                  'tasks': [],
+                };
+              });
+        } else {
```
gbuenoandrade, line 130: Nit: return early pattern

src/main/webapp/panelStyle.js
```diff
@@ -103,9 +102,20 @@ function createTextListElement(inputElementId, listElementId) {
     inputElement.value = '';
     return;
   }
-  phrase = phrase.replace(',', '');
+  phrase = phrase.split(',').join('');
```
gbuenoandrade, line 105: Merge issue?

austinteshuba:
@austinteshuba Fix issue #152 here

src/main/webapp/panelStyle.js
austinteshuba: Yeah merge issue - thanks for catching this!

src/main/webapp/assignPanel.js
austinteshuba: Agreed thanks!

src/main/webapp/assignPanel.js
austinteshuba: Good idea - fixed

austinteshuba:
@gbuenoandrade 
> A few questions:
> 
> 1. Should emails sharing the same subject from different senders generate multiple tasks? (I think so)
>    1.1 Maybe the tasks titles could be something like `<subject> <sender>` in that case (or always)
> 2. What does the reset button do?
> 3. Should the user be presented with the main assign panel when they press "confirm" from settings?

1) Yes it does generate different tasks, and that's a good suggestion I'll get that updated. 
2) The reset button just reverts the tasks settings to the old settings (if the user had 10 days previously but changed it to 7, then realized they don't want to change the setting anymore, they can hit reset)
3) I'm not sure! I think so though, that sounds like a better UX. I'll get working on this.

src/main/webapp/assignPanel.js
```diff
@@ -90,24 +90,17 @@ function parseEmailIdFromTaskNotes(task) {
...
-  }
-  if (index < 0) {
+  const emailIdIndex = notes.lastIndexOf('emailId:');
+
+  if (emailIdIndex === -1) {
     console.log('No email ID found!');
     return null;
   }
-  return emailIdValue.split(':')[1];
+  return notes.substr(emailIdIndex).split(':')[1];
```
gbuenoandrade, line 99: nit: just `notes.substr(emailIdIndex + 'emailId:'.length)`?

src/main/webapp/assignPanel.js
```diff
@@ -268,5 +261,6 @@ function assignDisplayNextEmail() {
     return;
   }
 
-  subjectLineElement.innerText = actionableEmails[0].subject;
+  subjectLineElement.innerText =
+      `(From: ${actionableEmails[0].sender}) ${actionableEmails[0].subject}`;
```
gbuenoandrade, line 265: Thanks for looking into this!
Though, now it shows "Subject: From: ...". I wonder if we should change the "Subject" label, but I'm not particularly concerned.

src/main/webapp/assignPanel.js
```diff
@@ -0,0 +1,266 @@
...
+ * Assumes that the email ID will be in the last line of the notes.
+ *
+ * @param {Task} task task object
+ * @return {string|null} id if present, null otherwise
+ */
+function parseEmailIdFromTaskNotes(task) {
+  // Get emailId from end of message
+  const notes = task.notes;
+  const minimumLength = 'emailId:'.length;
+  if (notes == null || notes.length < minimumLength) {
```
gbuenoandrade, line 89: nit: you could remove the `notes.length < minimumLength` check

src/main/webapp/assignPanel.js
austinteshuba: Yes that would work! Thanks!

src/main/webapp/assignPanel.js
austinteshuba: I think it should be fine, as it now looks like: Subject: (From: Austin Teshuba) ACTION REQUIRED...

I see what your saying though, it's just that space is a premium. I can't think of anything better right now, so I'll leave it for now and we can change this later (it'll be a 1 line PR)

src/main/webapp/assignPanel.js
austinteshuba: Agreed! Thanks!

PR #138, opened 2020-08-02
Authentication refactor to get user email

austinteshuba:
This will be used for the Assign panels' prioritization algorithm (I need the user's email so I can check if it's included in the 'TO' header. If it isn't, the email was likely sent as a result of a mailing list)


src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -134,41 +176,53 @@ public void doPost(
...
     } catch (CookieParseException e) {
       throw new CredentialVerificationException("idToken is not present / cannot be parsed!", e);
     }
 
+    String idToken = idTokenCookie.getValue();
+
+    return authenticationVerifier
+        .getUserEmail(idToken)
+        .orElseThrow(
+            () ->
```
tttzach, line 202: Does it work without this line?

src/test/java/AuthenticatedServletTestBase.java
```diff
@@ -43,8 +44,8 @@ public void setUp() throws Exception {
     authenticationVerifier = Mockito.mock(AuthenticationVerifier.class);
 
     // Authentication will always pass
-    Mockito.when(authenticationVerifier.verifyUserToken(ID_TOKEN_VALUE))
-        .thenReturn(AUTHENTICATION_VERIFIED);
+    Mockito.when(authenticationVerifier.getUserEmail(ID_TOKEN_VALUE))
+        .thenReturn(java.util.Optional.of(USER_EMAIL));
```
tttzach, line 48: Maybe add `import java.util.Optional`

src/main/java/com/google/sps/model/AuthenticationVerifierImpl.java
```diff
@@ -23,11 +23,13 @@
 import java.io.IOException;
 import java.security.GeneralSecurityException;
 import java.util.Collections;
+import java.util.Optional;
 
 /** Verifies authentication information with Google */
 public class AuthenticationVerifierImpl implements AuthenticationVerifier {
   @Override
-  public boolean verifyUserToken(String idToken) throws GeneralSecurityException, IOException {
+  public Optional<String> getUserEmail(String idToken)
```
tttzach, line 31: Interesting data structure - I did not know this existed

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
```diff
@@ -94,15 +95,36 @@ public final void doPost(HttpServletRequest request, HttpServletResponse respons
...
+   * @param googleCredential valid, verified google credential object
+   * @param userEmail the email address of the user
+   * @throws IOException if a read/write issue arises while processing the request
+   * @throws ServletException if the request cannot be handled due to unexpected errors
+   */
+  public void doGet(
+      HttpServletRequest request,
+      HttpServletResponse response,
+      Credential googleCredential,
+      String userEmail)
```
tttzach, line 123: Just curious - where do you use this parameter?

austinteshuba: It's not in this PR - it's a dependent PR (PR #139) as this is needed to check if the user was personally sent the email or if it was sent as a part of a mailing list

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: Hmm - I'm not sure which line you mean but the `.orElseThrow` unwraps the optional and throws an error if the value is null. This is required for proper error handling here

src/main/java/com/google/sps/model/AuthenticationVerifierImpl.java
austinteshuba: Yeah! Anton told me about it and I instantly loved it - I didn't know these existed outside of Swift!

src/test/java/AuthenticatedServletTestBase.java
austinteshuba: Agreed! Thanks

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: Chatted offline about this being the line `() ->` - this is required as this method takes a supplier (unfortunately)

src/main/java/com/google/sps/model/AuthenticationVerifierImpl.java
```diff
@@ -41,6 +43,8 @@ public boolean verifyUserToken(String idToken) throws GeneralSecurityException,
     // If the userToken is not null, the identity is verified and vice versa
     GoogleIdToken userToken = verifier.verify(idToken);
 
-    return userToken != null;
+    return userToken != null
+        ? Optional.ofNullable(userToken.getPayload().getEmail())
+        : Optional.empty();
```
antmar, line 48: There is an even cleaner way to do this with optionals:
```java
return Optional.ofNullable(userToken).map(t -> t.getPayload().getEmail())
```

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
antmar: This makes sense, because exceptions are used in exceptional circumstances, and you don't want to incur the overhead of creating an exception object every time.

src/main/java/com/google/sps/model/AuthenticationVerifierImpl.java
antmar: Nice use of Optional - I like how it help make the tests more readable too.

src/main/java/com/google/sps/model/AuthenticatedHttpServlet.java
austinteshuba: @antmar Makes sense - interesting!

src/main/java/com/google/sps/model/AuthenticationVerifierImpl.java
austinteshuba: Nifty! Fixed

PR #139, opened 2020-08-02
Create Prioritization system for ActionableMessages

austinteshuba:
The system assigns one of three priorities to a message: Low, medium, and high. High priority is assigned if the user starred the email. Medium priority is assigned if one of the following is true: the email was not sent by a mailing list or the email was marked important automatically by Gmail. An email is classified to be "sent from a mailing list" if it is either 1) not addressed directly to the current user or 2) sent to more than some N number of recipients (I set N to 15 here, but we can move this up or down if anyone has strong feelings). A low priority is assigned if the email is not of high or medium priority. 

Emails are returned to the server in order of priority, from highest to lowest, and then in order of date received (later it was received, the higher the priority).

This prioritization is quite rigid and isn't going to assign the perfect, real-world priority every time (if we were able to do that, we would assign priority via some discrete variable and sort by that, since the priority is absolute). However, this hopefully does enough of the sorting that the order emails are presented to the user makes sense.

Dependent on PR #155


src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
```diff
@@ -0,0 +1,117 @@
...
+import com.google.sps.exceptions.GmailMessageFormatException;
+import com.google.sps.utility.GmailUtility;
+import java.util.Arrays;
+import java.util.List;
+
+/** Contains methods that help create the non-trivial fields in an ActionableMessage. */
+public final class ActionableMessageHelperImpl implements ActionableMessageHelper {
+  // Public for testing purposes
+  public static final String UNDISCLOSED_RECIPIENTS_TO_HEADER = "undisclosed-recipients:;";
+  public static final int MAILING_LIST_LOWER_BOUND = 15;
```
tttzach, line 28: Did you mean the class is public for testing purposes?

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
```diff
@@ -0,0 +1,117 @@
...
+      Message message, String userEmail) {
+    if (isStarred(message)) {
+      return ActionableMessage.MessagePriority.HIGH;
+    }
+
+    if (!isFromMailingList(message, userEmail) || isImportant(message)) {
+      return ActionableMessage.MessagePriority.MEDIUM;
+    }
+
+    return ActionableMessage.MessagePriority.LOW;
```
tttzach, line 41: Nice use of early return pattern!

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    List<String> recipientHeaders = Arrays.asList(headerValue.split(","));
+    if (recipientHeaders.size() >= MAILING_LIST_LOWER_BOUND) {
+      return true;
+    }
+
+    boolean userEmailInHeader =
+        recipientHeaders.stream().filter((header) -> userEmailInHeader(header, userEmail)).count()
+            == 1;
```
tttzach, line 85: Maybe you could use `.allMatch` for this. Here's a Stack Overflow [link](https://stackoverflow.com/questions/35282418/use-streams-to-return-boolean-if-all-the-list-values-in-a-map-are-empty-not-empt) showing how to use that.

src/test/java/GmailActionableEmailsServletTest.java
```diff
@@ -150,21 +190,45 @@ public void nDaysZero() throws Exception {
...
 
   @Test
-  public void validResponse() throws Exception {
+  public void validResponseSortByDate() throws Exception {
     Mockito.when(request.getParameter("subjectLinePhrases"))
         .thenReturn(SUBJECT_LINE_PHRASES_STRING);
     Mockito.when(request.getParameter("unreadOnly")).thenReturn(String.valueOf(true));
     Mockito.when(request.getParameter("nDays")).thenReturn(String.valueOf(DEFAULT_N_DAYS));
+    Mockito.when(
+            actionableMessageHelper.assignMessagePriority(Mockito.any(), Mockito.eq(USER_EMAIL)))
```
tttzach, line 199: Is it possible to narrow down `Mockito.any()` any further?

src/test/java/GmailActionableEmailsServletTest.java
```diff
@@ -49,41 +51,79 @@
...
+  private static final Message messageHighPriority =
+      new Message()
+          .setId(MESSAGE_ID_LOW_PRIORITY)
+          .setPayload(new MessagePart().setHeaders(Collections.singletonList(subjectHeaderTwo)))
+          .setInternalDate(INTERNAL_DATE_LATE);
+
+  private static final List<Message> messagesOldestToNewest =
+      ImmutableList.of(messageOld, messageNew);
+  private static final List<Message> messagesPriorityLowestToHighest =
+      ImmutableList.of(messageLowPriority, messageHighPriority);
```
tttzach, line 96: These variables should be in constant case `MESSAGES_OLDEST_TO_NEWEST` and `MESSAGES_PRIORITY_LOWEST_TO_HIGHEST`

src/main/java/com/google/sps/model/ActionableMessage.java
```diff
@@ -63,4 +67,38 @@ public boolean equals(Object obj) {
...
+   * considered to be higher priority
+   *
+   * @param actionableMessage the ActionableMessage that this object is being compared against
+   * @return A positive integer if the current object has the higher priority, a negative integer if
+   *     the passed object has the higher priority, 0 if both objects have the same priority
+   */
+  @Override
+  public int compareTo(ActionableMessage actionableMessage) {
+    if (priority.getPriorityValue() == actionableMessage.priority.getPriorityValue()) {
+      return (int) (internalDate - actionableMessage.internalDate);
```
tttzach, line 83: Is it possible to get away with not casting the difference in `internalDates` to `int` and keeping them as `long`?

austinteshuba: Unfortunately no, since `compareTo` takes an integer as a return value. It's pretty cool how this works - if you return a positive integer, it means the current object is larger. If you return negative, the passed in object is larger. If you return 0, they are equal. 

The line you are referring to is just shorthand for "the email sent later is considered "larger" for sorting purposes"

src/test/java/GmailActionableEmailsServletTest.java
austinteshuba: Agreed! Thanks!

src/test/java/GmailActionableEmailsServletTest.java
austinteshuba: Maybe, but I almost want to keep this for clarity. This unit test does not consider priority - I want to make it clear that any passed in message will just have the default priority, so sorting will be only based on the secondary factor, `internalDate`

austinteshuba: @antmar @gbuenoandrade let me know your thoughts on this or if you have any suggestions!

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
gbuenoandrade: +1
RESOLVED

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
```diff
@@ -0,0 +1,117 @@
...
+   * @param message Message object to be checked
+   * @return true if starred, false otherwise
+   */
+  private boolean isStarred(Message message) {
+    List<String> labelIds = message.getLabelIds();
+    return labelIds != null && labelIds.contains("STARRED");
+  }
+
+  /**
+   * If the user's email is not in the "To" header, or if there are more than 15 recipients, the
```
gbuenoandrade, line 56: nit: `MAILING_LIST_LOWER_BOUND` instead of `15`

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
gbuenoandrade: Would `headerValue.contains(userEmail)` be enough? (or the `countMatches` method mentioned below if the number of occurrences is relevant)

If you go that way, it might be better to use [StringUtils.countMatches](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#countMatches-java.lang.CharSequence-char-) on line 78

src/main/java/com/google/sps/model/ActionableMessageHelper.java
```diff
@@ -0,0 +1,39 @@
...
+   * message where both of these are false. Finally, if a message was sent later than another
+   * message, it is considered to be of higher priority.
+   *
+   * @param message Message object to be assigned. Should be of type METADATA (with the "TO" header)
+   *     or FULL
+   * @param userEmail the email address of the current user. Used to check if the email was actually
+   *     sent to them or sent as a part of a mailing list
+   * @return 1 if messageOne has the higher priority, -1 is messageTwo has the higher priority 0 if
+   *     they have the same priority (very very rare - needs same internal date which is measured in
+   *     milliseconds)
```
gbuenoandrade, line 34: Doesn't this method return an absolute priority?

src/test/java/GmailActionableEmailsServletTest.java
gbuenoandrade: sgtm

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
austinteshuba: No, the properties themselves are public. These are the same values that should be used in the tests as well

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
austinteshuba: Thanks for pointing this out! Fixed this up using a new stream method I found, `anyMatch`.

src/test/java/GmailActionableEmailsServletTest.java
austinteshuba: RESOLVED

src/test/java/GmailActionableEmailsServletTest.java
austinteshuba: RESOLVED

src/main/java/com/google/sps/model/ActionableMessage.java
```diff
@@ -21,19 +21,23 @@
...
   private final String id;
   private final String subject;
+  private final long internalDate;
+  private final MessagePriority priority;
 
   /**
    * Creates an ActionableMessage
    *
    * @param id the message id of the Message in the user's Gmail account
    * @param subject the subject line of the message from the user's Gmail account
```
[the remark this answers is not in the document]
austinteshuba: Thanks! Fixed!

src/main/java/com/google/sps/model/ActionableMessage.java
```diff
@@ -44,9 +48,22 @@ public String getSubject() {
...
+  }
+
+  public MessagePriority getPriority() {
+    return priority;
+  }
+
   @Override
   public int hashCode() {
-    return new HashCodeBuilder(7, 17).append(id).append(subject).toHashCode();
+    return new HashCodeBuilder(7, 17)
```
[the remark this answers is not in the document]
austinteshuba: I suppose not! Thanks for pointing this out - fixed

src/main/java/com/google/sps/model/ActionableMessageHelper.java
austinteshuba: Ah oops - yes it does. Fixing!

src/main/java/com/google/sps/model/ActionableMessageHelper.java
```diff
@@ -0,0 +1,37 @@
...
+import com.google.api.services.gmail.model.Message;
+
+/** Contains methods that help create the non-trivial fields in an ActionableMessage */
+public interface ActionableMessageHelper {
+  /**
+   * Priority is determined by a series of arbitrary factors. If a message is starred by a user, it
+   * is considered high priority. Then, if a message is not from a mailing list or marked important,
+   * it is marked medium priority Otherwise, the message is of low priority. higher priority than a
+   * message where both of these are false. Finally, if a message was sent later than another
+   * message, it is considered to be of higher priority.
```
gbuenoandrade, line 26: "Priority is determined by a series of arbitrary factors. If a message is starred by a user, it is considered high priority. Then, if a message is not from a mailing list or marked important, it is marked medium priority Otherwise, the message is of low priority."

nit: I think the doc could stop there

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
```diff
@@ -0,0 +1,99 @@
...
+   *     sent to them or sent as a part of a mailing list
+   * @return true if from mailing list, false otherwise
+   * @throws GmailMessageFormatException if the message does not contain a "To" header (which is a
+   *     mandatory header, indicating the Message object is not of the correct type [METADATA with
+   *     correct header settings, or FULL])
+   */
+  private boolean isFromMailingList(Message message, String userEmail) {
+    MessagePartHeader toHeader = GmailUtility.extractHeader(message, "To");
+    if (toHeader == null) {
+      throw new GmailMessageFormatException("'To' header not present. Check message format");
```
gbuenoandrade, line 72: nit: append the message to the exception message?

src/main/java/com/google/sps/model/ActionableMessageHelper.java
austinteshuba: Agreed - thanks!

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
austinteshuba: Good Idea! Thanks

austinteshuba: Good idea - I'll append the messageID (since that's the only 'guaranteed' property unless if the passed message is null)

src/main/java/com/google/sps/model/ActionableMessage.java
```diff
@@ -68,7 +89,43 @@ public boolean equals(Object obj) {
...
+  @Override
+  public int compareTo(ActionableMessage actionableMessage) {
+    if (priority.getPriorityValue() == actionableMessage.priority.getPriorityValue()) {
+      return (int) (internalDate - actionableMessage.internalDate);
+    }
+
+    return priority.getPriorityValue() - actionableMessage.priority.getPriorityValue();
+  }
+
+  public enum MessagePriority {
```
antmar, line 116: Nice use of custom enum classes!

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
```diff
@@ -0,0 +1,100 @@
...
+  private boolean isFromMailingList(Message message, String userEmail) {
+    MessagePartHeader toHeader = GmailUtility.extractHeader(message, "To");
+    if (toHeader == null) {
+      throw new GmailMessageFormatException(
+          "'To' header not present. Check message format. Message ID:" + message.getId());
+    }
+    String headerValue = toHeader.getValue();
+    if (headerValue.equals(UNDISCLOSED_RECIPIENTS_TO_HEADER)) {
+      return true;
+    }
```
antmar, line 78: This check seems redundant, because the check `userEmailInHeader` will be false for undisclosed recipients.

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
```diff
@@ -0,0 +1,100 @@
...
+   *
+   * @param message Message object to be checked
+   * @param userEmail the email address of the current user. Used to check if the email was actually
+   *     sent to them or sent as a part of a mailing list
+   * @return true if from mailing list, false otherwise
+   * @throws GmailMessageFormatException if the message does not contain a "To" header (which is a
+   *     mandatory header, indicating the Message object is not of the correct type [METADATA with
+   *     correct header settings, or FULL])
+   */
+  private boolean isFromMailingList(Message message, String userEmail) {
```
antmar, line 69: I think this method should be named something more general like `isBulkMail` or `isGroupMail`. I think this is a good prioritization criteria, but it's a lot more general than just "sent from a mailing list".

If you want to check for an actual mailing list, there are headers included by most mailing list software that will tell you it's from a mailing list, and even help identify the precedence/priority: https://stackoverflow.com/q/6378773

src/main/java/com/google/sps/servlets/GmailActionableEmailsServlet.java
```diff
@@ -113,7 +126,8 @@ public void doGet(
     List<ActionableMessage> actionableEmails =
         gmailClient.getActionableEmails(subjectLinePhrases, unreadOnly, nDays, METADATA_HEADERS)
             .stream()
-            .map(this::createActionableMessage)
+            .map((message) -> createActionableMessage(message, userEmail))
+            .sorted(Comparator.reverseOrder())
```
antmar, line 130: Nice use of `sorted` in stream processing!

src/test/java/ActionableMessageHelperImplTest.java
```diff
@@ -0,0 +1,138 @@
...
+  private static final MessagePartHeader toHeaderNotUserEmail =
+      new MessagePartHeader().setName("To").setValue(TO_HEADER_VALUE_NOT_USER_EMAIL);
+  private static final MessagePartHeader toHeaderUndisclosedRecipients =
+      new MessagePartHeader().setName("To").setValue(TO_HEADER_VALUE_UNDISCLOSED_RECIPIENTS);
+  private static final MessagePartHeader toHeaderTooManyRecipients =
+      new MessagePartHeader().setName("To").setValue(TO_HEADER_VALUE_TOO_MANY_RECIPIENTS);
+
+  private static final Message starredMessage =
+      new Message()
+          .setPayload(new MessagePart().setHeaders(Collections.singletonList(toHeaderNotUserEmail)))
```
antmar, line 59: Very minor comment: `ImmutableList.of(element)` is a way to express a single-item list with less typing. I think the shorter form is actually more readable, because more people are familiar with `ImmutableList.of` construct.

src/test/java/ActionableMessageHelperImplTest.java
```diff
@@ -0,0 +1,138 @@
...
+      new Message()
+          .setPayload(
+              new MessagePart().setHeaders(Collections.singletonList(toHeaderNotUserEmail)));
+  private static final Message userEmailMessage =
+      new Message()
+          .setPayload(new MessagePart().setHeaders(Collections.singletonList(toHeaderUserEmail)));
+  private static final Message messageWithoutToHeader = new Message();
+
+  private static String createHeaderWithNRecipients(int n, String headerValue, String delimiter) {
+    StringBuilder stringBuilder = new StringBuilder();
```
antmar, line 84: Good choice of using StringBuilder over string concatenation!

src/test/java/ActionableMessageHelperImplTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+  @Test
+  public void getPriorityStarredMessage() {
+    ActionableMessage.MessagePriority actualPriority =
+        actionableMessageHelperImpl.assignMessagePriority(starredMessage, USER_EMAIL);
+    Assert.assertEquals(ActionableMessage.MessagePriority.HIGH, actualPriority);
+  }
+
+  @Test
+  public void getPriorityImportantMessage() {
```
antmar, line 105: What about important and starred? Other combinations to check precedence?

src/test/java/GmailActionableEmailsServletTest.java
```diff
@@ -46,10 +48,12 @@
...
 
-  private static final String MESSAGE_ID_ONE = "messageOne";
-  private static final String MESSAGE_ID_TWO = "messageTwo";
+  private static final String MESSAGE_ID_OLD = "messageOne";
+  private static final String MESSAGE_ID_NEW = "messageTwo";
+  private static final String MESSAGE_ID_LOW_PRIORITY = "messageTwo";
+  private static final String MESSAGE_ID_HIGH_PRIORITY = "messageTwo";
   private static final String SUBJECT_VALUE_ONE = "subjectValueOne";
   private static final String SUBJECT_VALUE_TWO = "subjectValueTwo";
   private static final String SENDER_EMAIl = "example@example.com";
```
antmar, line 59: Nit: the last letter should be capital 'L'

src/main/java/com/google/sps/model/ActionableMessage.java
austinteshuba: Thank you!

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
austinteshuba: I agree - I'll change up the name

I also added a check looking for a `List-ID` header, which (if present) means the email is likely from a mailing list. Thanks for sharing this!

src/main/java/com/google/sps/servlets/GmailActionableEmailsServlet.java
austinteshuba: Thank you!

src/test/java/ActionableMessageHelperImplTest.java
austinteshuba: Thank you!

src/test/java/ActionableMessageHelperImplTest.java
austinteshuba: 100% Agreed - fixing this now! Thanks

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
austinteshuba: Agreed - thanks!

src/test/java/ActionableMessageHelperImplTest.java
austinteshuba: Fixed! Thanks for the suggestion

src/test/java/GmailActionableEmailsServletTest.java
austinteshuba: Fixed! Good catch

src/main/java/com/google/sps/model/ActionableMessageHelperImpl.java
austinteshuba: Agreed - thanks!

PR #140, opened 2020-08-03
Default values for gmail and tasks

austinteshuba:
This implements default values for the Gmail and Tasks panel while it is loading. 

Add Calendar to this before merging to master (unless if it handled in PR #99)


src/main/webapp/populateData.js
```diff
@@ -45,8 +45,14 @@ function populateGmail() {
...
   const mHours = 3;
 
+  // Set known values and placeholders while loading
   nDaysContainer.innerText = nDays;
   mHoursContainer.innerText = mHours;
+  unreadEmailsContainer.innerText = '-';
+  unreadEmailsThreeHrsContainer.innerText = '-';
+  importantEmailsContainer.innerText = '-';
+  senderInitialContainer.innerText = '-';
+  senderContainer.innerText = '-';
```
gbuenoandrade, line 55: I wonder if `...` would be more clear (random suggestion, not really based on any UX data)

src/main/webapp/index.html
```diff
@@ -94,20 +94,18 @@ <h3 class="panel__section-heading">Tasks</h3>
             </div>
             <div class="panel__content-entry">
               <img class="panel__content-icon" alt="clipboard-icon" src="assets/clipboard.png">
-              <p class="panel__text-large" id="tasks-due-today">4 due today
+              <p class="panel__text-large"><span id="tasks-due-today">4</span> due today
```
antmar, line 97: Consider replacing the placeholders in the HTML as well. Otherwise, I think there will be a period of time when the hard-coded values are shown, probably while the page is authenticating.

src/main/webapp/populateData.js
austinteshuba: Agreed! Thanks

src/main/webapp/index.html
austinteshuba: You're absolutely right - fixed!

PR #141, opened 2020-08-03
Add settings for Gmail panel

austinteshuba:
Create a settings panel for the Gmail panel, so nDays and mHours can be selected by the user. 

This is implemented with the same pattern and logic used by the assign panel (PR #125).

Settings Panel View: 
Screenshot 2020-08-02 at 11 15 59 PM

Panel View: 
Screenshot 2020-08-02 at 11 16 36 PM

Dependent on PR #125

Closes Issue #83


src/main/webapp/gmailPanel.js
```diff
@@ -0,0 +1,99 @@
...
+  // Get values for nDays and mHours
+  const nDaysSettingsContainer = document.querySelector('#gmailSettingsNDays');
+  const mHoursSettingsContainer =
+      document.querySelector('#gmailSettingsMHours');
+  gmailNDays = parseInt(nDaysSettingsContainer.innerText);
+  gmailMHours = parseInt(mHoursSettingsContainer.innerText);
+
+  // Get containers for all gmail fields
+  const nDaysContainer = document.querySelector('#gmailNDays');
+  const mHoursContainer = document.querySelector('#gmailMHours');
```
tttzach, line 36: I don't know if [this](https://google.github.io/styleguide/htmlcssguide.html#ID_and_Class_Name_Delimiters) applies to HTML IDs as well, I believe these would be better - separated. Same everywhere else if you plan on changing this.

src/main/webapp/gmailPanel.js
```diff
@@ -0,0 +1,99 @@
...
+          signOut();
+        }
+      });
+}
+
+/**
+ * Reset the values for nDays and mHours in the settings panel
+ */
+function gmailRevertSettings() {
+  // Get values for nDays and mHours
```
tttzach, line 93: nit: I think this comment contradicts the javadoc a little, you can remove it

src/main/webapp/gmailPanel.js
austinteshuba: Oh interesting I didn't know that was in style guide. I'll fix this everywhere

src/main/webapp/gmailPanel.js
austinteshuba: Agreed - thanks!

src/main/webapp/index.html
```diff
@@ -44,37 +46,62 @@ <h2 class="panel-row__header u-align-with-first-panel u-text-shadow">Services (a
...
-            <div class="panel__content-entry">
-              <div class="panel__round-letter">
-                <p class="panel__text-data-point u-text-white" id="gmailSenderInitial">
-              </div>
-              <p class="panel__text-large"><span id="gmailSender"></span> sent the most unread emails
+          </div>
+          <div class="panel__content-entry">
+            <p class="panel__text-large">Search <span class="u-bold-underline" id="gmail-settings-m-hours">3</span> hours of recent email
+            <div class="panel__two-buttons-incremental u-right-align">
+              <a class="panel__button-incremental" onclick="incrementElement('gmail-settings-m-hours')"><span>+</span></a>
```
gbuenoandrade, line 95: Does the increment have a limit? Would it make sense to add an extra day when it reaches 24?

src/main/webapp/index.html
```diff
@@ -44,37 +46,62 @@ <h2 class="panel-row__header u-align-with-first-panel u-text-shadow">Services (a
...
+            <p class="panel__text-large"><span id="gmail-sender"></span> sent the most unread emails
+          </div>
+        </div>
+        <!-- Settings section -->
+        <h3 hidden class="panel__section-heading u-remove-top-margin gmail-settings">Settings:</h3>
+        <div hidden class="panel__text-body gmail-settings">
+          <div class="panel__content-entry u-remove-top-margin">
+            <p class="panel__text-large">Search <span class="u-bold-underline" id="gmail-settings-n-days">7</span> days of email
+            <div class="panel__two-buttons-incremental u-right-align">
+              <a class="panel__button-incremental" onclick="incrementElement('gmail-settings-n-days')"><span>+</span></a>
```
gbuenoandrade, line 88: Is the increment limited to 30? What do you think of making it "infinite" if the user tries to go beyond the limit?

I think it would also be cool if the "+" disappeared when the user reached "infinity" or the limit depending on what you decide (same for "-"). (This only makes sense if it's easy to implement.)

src/main/webapp/index.html
austinteshuba: Yes the limit is 30 - and I actually think maybe the limit on this "hours" property should be 23. The whole point is to see how many recent emails there are, and by allowing "30" hours it's defeating the purpose (plus that's an awkward amount of time). Fixing

src/main/webapp/index.html
austinteshuba: I think that we should have a limit for performance reasons. 

Batching should help with this, but I know power users (like me) have a **lot** of email. I think I have roughly 100,000 unread emails in my Gmail. I'm worried that if we remove the cap, the system will not perform well even with the best methods implemented. 

I wanted the cap to be generous, and make semantic sense, so I thought 30 would be a good idea - let me know if you think a different limit would be better (I just picked this arbitrarily).

In terms of adding and removing the +, I can toggle the disabled feature on the button which may work out nicely, but that's going to take some time (~1 hour). If I have some time tomorrow, I'll take a look into it. Good idea 👍

austinteshuba: I did it!

src/main/webapp/index.html
gbuenoandrade: Sgtm

src/main/webapp/index.html
gbuenoandrade: Cool, looks great!

Regarding the cap, I think it would be nice to give power users the "power" to parse all their emails if they wanted, but that's totally minor. I think the +/- fading out you did is pretty cool and wraps up this discussion in a great way!

austinteshuba: Thanks! 

RESOLVED

PR #142, opened 2020-08-03
Use batching to more efficiently handle large volumes of Gmail requests

austinteshuba:
Use batching to be able to 1) list user messages in excess of 100 (the Gmail API limit per request) and 2) be able to retrieve the contents of message objects up to 100x faster (batch requests can retrieve 100 gmail message bodies at once).

This should not break anyone else's code - changes were made within existing infrastructure. 

Large volume example: 
Screenshot 2020-08-03 at 12 19 20 AM

Dependent on PR #112


src/main/java/com/google/sps/model/GmailClientImpl.java
```diff
@@ -113,21 +145,39 @@ public Message getUserMessageWithMetadataHeaders(String messageId, List<String>
...
+            .users()
+            .messages()
+            .get("me", currentMessage.getId())
+            .setFormat(messageFormat.formatValue)
+            .queue(batchRequest, batchCallback);
+        messageIndex++;
+      }
+
+      batchRequest.execute();
+    }
```
tttzach, line 173: Interesting!

src/main/java/com/google/sps/model/GmailClientImpl.java
```diff
@@ -137,16 +187,57 @@ public Message getUserMessageWithMetadataHeaders(String messageId, List<String>
...
+            .messages()
+            .get("me", currentMessage.getId())
+            .setFormat(MessageFormat.METADATA.formatValue)
+            .setMetadataHeaders(metadataHeaders)
+            .queue(batchRequest, batchCallback);
+        messageIndex++;
+      }
+
+      batchRequest.execute();
+    }
```
tttzach, line 216: At a glance this looks the same as the loop in `listUserMessagesWithFormat`. Would it be possible to abstract this out?

austinteshuba: Yeah I think I figured this out!

src/main/java/com/google/sps/model/GmailClientImpl.java
austinteshuba: Thanks!

src/main/java/com/google/sps/model/GmailClientImpl.java
```diff
@@ -42,13 +48,42 @@ private GmailClientImpl(Credential credential) {
...
+   */
   @Override
   public List<Message> listUserMessages(String query) throws IOException {
-    // Null if no messages present. Convert to empty list for ease
-    List<Message> messages =
-        gmailService.users().messages().list("me").setQ(query).execute().getMessages();
-
-    return messages != null ? messages : new ArrayList<>();
+    List<Message> userMessages = new ArrayList<>();
+    List<Message> newBatchUserMessages;
```
antmar, line 65: You can move this variable definition inside the do-while loop.

src/main/java/com/google/sps/model/GmailClientImpl.java
```diff
@@ -42,13 +48,42 @@ private GmailClientImpl(Credential credential) {
             .build();
   }
 
+  /**
+   * List the messages in a user's Gmail account that match the passed query. Results are returned
+   * in batches of 100 from the Gmail API, though a next page token is included if another batch
+   * still exists. This method will use these tokens to get all of the emails that match the query,
+   * even if that is in excess of 100.
```
antmar, line 55: It's weird to have javadoc on a method override, because the external API of the method should be identical to the original declaration. Consider adding a comment with these implementation details inside the method body.

src/main/java/com/google/sps/model/GmailClientImpl.java
austinteshuba: Agreed! Done!

src/main/java/com/google/sps/model/GmailClientImpl.java
austinteshuba: Good point - Thanks for this! Fixed

PR #143, opened 2020-08-04
Refactor Directions for Go panel

tttzach:
Remove Directions servlet which is to be replaced by the Go servlet and refactor to include more methods for the Go panel.


src/test/java/DirectionsClientTest.java
```diff
@@ -33,50 +29,48 @@
...
-        ImmutableList.of(
-            A_TO_B_NO_WAYPOINTS
-                .get(0)
-                .replace(
-                    "(45,-73 -> 43,-80)",
-                    "(45.00000000,-73.00000000 -> 43.00000000,-80.00000000)"));
+    long actual = DirectionsClient.getTotalTravelTime(directionsResult);
 
-    Assert.assertEquals(expected, actual);
+    Assert.assertEquals(18, actual);
```
austinteshuba, line 74: I would use `durationOne.inSeconds + durationTwo.inSeconds` here for clarity

tttzach: I agree, 18 is confusing. Thanks!

tttzach: `RESOLVED`

PR #145, opened 2020-08-04
Refactor Places for Go panel

tttzach:
Refactor Places for Go panel.


src/main/java/com/google/sps/model/PlacesClient.java
```diff
@@ -38,18 +35,18 @@
...
    */
-  public static List<String> getFormattedAddresses(PlacesSearchResponse response) {
-    return Arrays.asList(response.results).stream()
-        .map(result -> result.formattedAddress)
-        .collect(Collectors.toList());
+  public static String getPlaceId(PlacesSearchResponse response) {
+    if (response.results.length != 0) {
+      return response.results[0].placeId;
+    }
+    return null;
```
austinteshuba, line 50: Use an Optional is returning null is a possibility

tttzach: That's a good point, I've changed this to be an empty string instead of Optional because it doesn't really matter if it's empty, just need a way to filter and empty string does the trick. Thanks!

src/main/java/com/google/sps/model/PlacesClient.java
```diff
@@ -38,18 +35,18 @@
    * @throws PlacesException A custom exception is thrown to signal an error pertaining to the
    *     Places API.
    */
-  List<String> searchNearby(LatLng location, PlaceType placeType, RankBy rankBy)
-      throws PlacesException;
+  String searchNearby(LatLng location, PlaceType placeType, RankBy rankBy) throws PlacesException;
```
antmar, line 38: Update the method documentation.

Also, what if there is more than one result - document that you pick the first one.

src/main/java/com/google/sps/model/PlacesClient.java
```diff
@@ -38,18 +35,18 @@
...
+   * public for testing purposes.
    *
-   * @param response The PlacesSeachResponse object to get formatted addresses from
+   * @param response The PlacesSearchResponse object to get formatted addresses from
    */
-  public static List<String> getFormattedAddresses(PlacesSearchResponse response) {
-    return Arrays.asList(response.results).stream()
-        .map(result -> result.formattedAddress)
-        .collect(Collectors.toList());
+  public static String getPlaceId(PlacesSearchResponse response) {
```
antmar, line 46: If `getPlaceId` will only be used in `PlacesClientImpl`, then just move it there. You could still leave it public for testing.

Alternatively, if it's used in other classes, then I suggest moving it to a separate utility class.

src/main/java/com/google/sps/model/PlacesClient.java
tttzach: Done, thanks! I should reread all the documentation and change them to be more accurate.

src/main/java/com/google/sps/model/PlacesClient.java
tttzach: Sounds like a plan, thanks!

src/test/java/PlacesClientTest.java
```diff
@@ -25,53 +23,50 @@
 
 /**
  * Test that PlacesClientImpl correctly converts a PlacesSearchResponse object obtained from the
- * Places API into a list of formatted addresses
+ * Places API into a list of Place IDs.
  */
 @RunWith(JUnit4.class)
-public class PlacesClientImplTest {
+public class PlacesClientTest {
```
antmar, line 29: Should this be `PlacesClientImplTest`, because you are testing the static method in the impl file?
(With appropriate file name change?)

src/main/java/com/google/sps/model/PlacesClient.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/PlacesClient.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/PlacesClient.java
tttzach: `RESOLVED`

src/test/java/PlacesClientTest.java
tttzach: Just moved this into PlacesResultUtility for consistency. Thanks for pointing that out!

tttzach: i.e. moved method to PlacesResultUtility and renamed this to PlacesResultUtilityTest

tttzach: `RESOLVED`

PR #146, opened 2020-08-04
Refactor Tasks for Go panel

tttzach:
Refactor Tasks for Go panel.


src/main/java/com/google/sps/model/TasksClientImpl.java
```diff
@@ -61,6 +62,17 @@ private TasksClientImpl(Credential credential) {
     return taskLists != null ? taskLists : new ArrayList<>();
   }
 
+  @Override
+  public List<TaskList> listTaskLists(List<String> taskListTitles) throws IOException {
+    // returns null if no taskLists exist. Convert to empty list for ease.
+    List<TaskList> taskLists = tasksService.tasklists().list().execute().getItems();
```
austinteshuba, line 68: Call `listTasklists()` here to avoid code duplication

src/main/java/com/google/sps/servlets/TasksServlet.java
```diff
@@ -212,7 +212,7 @@ private long countTasksOverdue(List<Task> tasks) {
    * @return List of tasks from all task lists in user's account
    * @throws IOException if an issue occurs with the tasksService
    */
-  private List<Task> getAllTasksFromAllTaskLists(TasksClient tasksClient) throws IOException {
+  public static List<Task> getAllTasksFromAllTaskLists(TasksClient tasksClient) throws IOException {
```
austinteshuba, line 215: If you also need this I agree this can be public and static, but this should probably be in the TasksClient

src/main/java/com/google/sps/model/TasksClientImpl.java
tttzach: Great point, thanks!

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: Agreed, this is used in the Go servlet as well so this would be perfect - not sure why I did not do this in the first place. Thanks!

src/main/java/com/google/sps/model/TasksClient.java
```diff
@@ -38,6 +39,14 @@
    */
   List<TaskList> listTaskLists() throws IOException;
 
+  /**
+   * Get all task lists in a user's Tasks account with the specified task list titles
```
gbuenoandrade, line 43: Can you mention how the titles should be matched? I.e., exact match? title must be contained?

src/main/java/com/google/sps/model/TasksClient.java
```diff
@@ -38,6 +39,14 @@
...
+   *
+   * @return List of task lists with any of the given task list titles
+   * @throws IOException if an issue occurs with the TasksService
+   */
+  List<TaskList> listTaskLists(List<String> taskListTitles) throws IOException;
```
gbuenoandrade, line 48: Nit: `Set<String>` is potentially faster (depending on the implementation), but, more importantly, it makes clearer that the order of the titles doesn't matter, it's just a set.

src/main/java/com/google/sps/model/TasksClient.java
```diff
@@ -56,4 +65,40 @@
...
+  /**
+   * Get the tasks in the user's task lists with the given task list IDs
+   *
+   * @param tasksClient Either a mock TaskClient or a taskClient with a valid credential
+   * @param taskListTitles List of task list IDs which tasks should be obtained from
+   * @return List of tasks from specified task lists in user's account
+   * @throws IOException if an issue occurs with the tasksService
+   */
+  static List<Task> getAllTasksFromSpecificTaskLists(
+      TasksClient tasksClient, List<String> taskListIds) throws IOException {
```
gbuenoandrade, line 94: Same as above regarding `Set`

src/main/java/com/google/sps/model/TasksClientImpl.java
```diff
@@ -61,6 +62,17 @@ private TasksClientImpl(Credential credential) {
     return taskLists != null ? taskLists : new ArrayList<>();
   }
 
+  @Override
+  public List<TaskList> listTaskLists(List<String> taskListTitles) throws IOException {
+    // returns null if no taskLists exist. Convert to empty list for ease.
+    List<TaskList> taskLists = listTaskLists();
+    taskLists = taskLists != null ? taskLists : new ArrayList<>();
```
gbuenoandrade, line 69: I think it's okay to trust that `listTaskLists()`won't return null (maybe add a comment to `listTaskLists` saying that it should not return null)

src/main/java/com/google/sps/model/TasksClient.java
tttzach: Moved some methods into TasksClient in another PR and hence this is not needed anymore so I removed it. Thanks and apologies you had to review that.

src/main/java/com/google/sps/model/TasksClient.java
tttzach: Removed - from above!

src/main/java/com/google/sps/model/TasksClient.java
tttzach: I have changed `taskListIds` to be a `Set<String>`, thanks!

src/main/java/com/google/sps/model/TasksClientImpl.java
tttzach: I have changed the comment, thanks!

src/main/java/com/google/sps/model/TasksClient.java
tttzach: Should I change `List<Task>` to `Set<Task>`?

tttzach:
@antmar Yeah you're right. The tests are in https://github.com/googleinterns/blueprint-step-2020/pull/132/files but that has been left to the end because it would result in a lot of overhead refactor including this PR. I assume it's not going to be merged in so I'll definitely move the tests here. Thanks for pointing that out! It wouldn't be difficult to move the static methods into utilities as well so i'll do that while moving the tests over.

tttzach:
> I think you need some tests for the public static methods you are adding. I thought that was the point of the refactoring?
> 
> Also, the usual comment that I think static methods should go into a separate utility instead of the interface itself. However, we haven't been consistent about it, so let's keep it as is.

@antmar Yeah you're right. They are tested in the servlet tests which are in https://github.com/googleinterns/blueprint-step-2020/pull/132/files but I agree there should be unit tests for the methods. Thanks for pointing that out! It wouldn't be difficult to move the static methods into utilities so i'll do that as well.

src/main/java/com/google/sps/model/TasksClient.java
gbuenoandrade: I just added a comment to your tests talking about the order. A `Set` would be neat if you don't want to give any guarantee regarding the order. (You could also use [Collection](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html) to make it more generic)

src/test/java/TasksUtilityTest.java
```diff
@@ -0,0 +1,102 @@
...
+    Assert.assertEquals(NO_TASKS, actual);
+  }
+
+  @Test
+  public void getSomeTasksFromSomeTaskLists() throws Exception {
+    Mockito.when(tasksClient.listTaskLists()).thenReturn(SOME_TASK_LISTS);
+    Mockito.when(tasksClient.listTasks(TASK_LIST_ONE)).thenReturn(ONE_TASK);
+    Mockito.when(tasksClient.listTasks(TASK_LIST_TWO)).thenReturn(SOME_TASKS);
+    List<Task> actual = TasksUtility.getAllTasksFromAllTaskLists(tasksClient);
+    Assert.assertEquals(ImmutableList.of(TASK_ONE, TASK_ONE, TASK_TWO), actual);
```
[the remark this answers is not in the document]
tttzach: Oh right, because we use Set now, the order might not be the same. Looks like I'd need to import [AssertJ](https://assertj.github.io/doc/) to do `assertThat(actual).containsExactlyInAnyOrder(expected);`. Would that be preferred over sorting both actual and expected?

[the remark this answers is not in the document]
tttzach: I just changed it to this, let me know if that works
```
Assert.assertTrue(expected.containsAll(actual));
Assert.assertTrue(actual.containsAll(expected));
```

src/test/java/TasksUtilityTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+  @Test
+  public void getAllTasksFromSelectedTaskList() throws Exception {
+    Mockito.when(tasksClient.listTaskLists()).thenReturn(SOME_TASK_LISTS);
+    Mockito.when(tasksClient.listTasks(TASK_LIST_ONE)).thenReturn(SOME_TASKS);
+    Mockito.when(tasksClient.listTasks(TASK_LIST_TWO)).thenReturn(SOME_TASKS);
+    List<Task> actual =
+        TasksUtility.getAllTasksFromSpecificTaskLists(
+            tasksClient, ImmutableSet.of(TASK_LIST_ID_ONE));
+    Assert.assertEquals(SOME_TASKS, actual);
```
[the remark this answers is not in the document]
tttzach: Got it!

src/test/java/TasksUtilityTest.java
```diff
@@ -0,0 +1,102 @@
...
+  @Before
+  public void setUp() {
+    tasksClient = Mockito.mock(TasksClient.class);
+  }
+
+  @Test
+  public void getAllTasksFromNoTaskLists() throws Exception {
+    Mockito.when(tasksClient.listTaskLists()).thenReturn(NO_TASK_LISTS);
+    List<Task> actual = TasksUtility.getAllTasksFromAllTaskLists(tasksClient);
+    Assert.assertEquals(ImmutableList.of(), actual);
```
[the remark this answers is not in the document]
tttzach: That's cleaner, thanks!

src/main/java/com/google/sps/model/TasksClientImpl.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/servlets/TasksServlet.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/TasksClient.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/TasksClient.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/TasksClientImpl.java
tttzach: `RESOLVED`

src/test/java/TasksUtilityTest.java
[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/test/java/TasksUtilityTest.java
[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/TasksClient.java
tttzach: Set is really neat, I looked a little deeper to see if I could change from List<Task> to Set<Task> but it might affect some of the other servlets so I don't think it might be a good idea at this time.

src/test/java/TasksUtilityTest.java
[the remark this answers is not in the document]
gbuenoandrade: > Oh right, because we use Set now, the order might not be the same
Not only because we are using a Set, but even before that, the method never "promised" any order.

Works for me, the double check was a neat idea!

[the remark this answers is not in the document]
tttzach: Oh true you're right, I should have considered that from the very start then. Thankfully the Tasks API does return tasks and tasklists in order every time. Thanks for pointing that out!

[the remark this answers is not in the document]
tttzach: `RESOLVED`

src/main/java/com/google/sps/model/TasksClient.java
tttzach: `RESOLVED`

PR #147, opened 2020-08-04
Add methods to Locations Utility for Go panel

tttzach:
Add methods to Locations utility for Go panel.


src/test/java/LocationsUtilityTest.java
```diff
@@ -118,4 +119,24 @@ public void getLocationAllTasks() {
         ImmutableList.of("Google Kitchener", "Google Kitchener", ""),
         LocationsUtility.getLocations(PREFIX, ALL_TASKS));
   }
+
+  @Test
```
austinteshuba, line 123: What happens in the null case (i.e. you pass an empty list?). Looks like there should be no issues, but it might be worth adding a test for as this is easy to mess up later

src/main/java/com/google/sps/utility/LocationsUtility.java
```diff
@@ -58,4 +59,27 @@ private static String getLocation(String prefix, String taskNotes) {
...
+  /**
+   * Generates all permutations of the lists recursively. Scope of method is public for testing
+   * purposes.
+   *
+   * @param lists A list of lists to generate permutations for.
+   * @param result A pointer to the result since the method is recursive.
+   * @param depth An int which specifies the list in lists to traverse.
+   * @param current A list which represents the result for the current recursive stack.
+   */
+  public static void generateCombinations(
```
austinteshuba, line 72: If you must use a pointer for the result, I would suggest creating a wrapper class that calls this recursive function and then returns the result. 

i.e. Call `generateCombinations(lists)`, which then calls this recursive function (which should then be private), and then returns `result`

src/test/java/LocationsUtilityTest.java
tttzach: Good point, edge case tests are important either way. Thanks!

src/main/java/com/google/sps/utility/LocationsUtility.java
tttzach: Wrapper functions, good idea! Thanks!

src/test/java/LocationsUtilityTest.java
```diff
@@ -106,4 +107,29 @@ public void getMultipleUniqueLocations() {
         LocationsUtility.getLocations(
             PREFIX, ImmutableList.of(TASK_WITH_LOCATION_ONE, TASK_WITH_LOCATION_TWO)));
   }
+
+  @Test
+  public void getCombinationsNoLists() {
+    List<List<String>> noLists = ImmutableList.of();
+    List<List<String>> actual = LocationsUtility.generateCombinations(noLists);
+    Assert.assertEquals(ImmutableList.of(ImmutableList.of()), actual);
```
gbuenoandrade, line 115: nit: `actual.isEmpty()`

src/main/java/com/google/sps/utility/LocationsUtility.java
```diff
@@ -63,4 +64,39 @@ private static String getLocation(Pattern pattern, String prefix, String taskNot
...
+   * purposes.
+   *
+   * @param lists A list of lists to generate permutations for.
+   */
+  public static List<List<String>> generateCombinations(List<List<String>> lists) {
+    List<List<String>> combinations = new ArrayList<>();
+    generateCombinationsHelper(lists, combinations, 0, new ArrayList<>());
+    return combinations;
+  }
 }
```
gbuenoandrade, line 102: I'm confused by what you mean by combination and permutation. When you say "Generates all permutations of the lists recursively", I would expect the method to return a list with lists.length! elements, where each element would be a possible permutation of the elements of `lists`. Could you please clarify?

Also, why is this being done?

src/test/java/LocationsUtilityTest.java
```diff
@@ -106,4 +107,29 @@ public void getMultipleUniqueLocations() {
...
+  }
+
+  @Test
+  public void getCombinationsOneList() {
+    List<List<String>> oneList = ImmutableList.of(ImmutableList.of("1", "2", "3"));
+    List<List<String>> actual = LocationsUtility.generateCombinations(oneList);
+    Assert.assertEquals(
+        ImmutableList.of(ImmutableList.of("1"), ImmutableList.of("2"), ImmutableList.of("3")),
+        actual);
```
gbuenoandrade, line 124: Please see my comment above. I don't understand why that's the expected output.

src/test/java/LocationsUtilityTest.java
```diff
@@ -106,4 +107,29 @@ public void getMultipleUniqueLocations() {
...
+  }
+
+  @Test
+  public void getCombinationsTwoLists() {
+    List<List<String>> twoLists =
+        ImmutableList.of(ImmutableList.of("1", "2"), ImmutableList.of("3"));
+    List<List<String>> actual = LocationsUtility.generateCombinations(twoLists);
+    Assert.assertEquals(
+        ImmutableList.of(ImmutableList.of("1", "3"), ImmutableList.of("2", "3")), actual);
```
gbuenoandrade, line 133: ditto

gbuenoandrade:
> Refactor Locations utility for Go panel.

I think this PR seems to be adding a feature instead of refactoring code.

src/main/java/com/google/sps/utility/LocationsUtility.java
gbuenoandrade: Do you mean all possible combinations containing a single element from each list?

tttzach: Oh I meant combinations for everything, thanks for catching that.

tttzach: Not permutations because order does not matter.

src/test/java/LocationsUtilityTest.java
tttzach: I can't do `actual.isEmpty()` for that because the nested list makes the list not empty but I managed to change this for two of the other tests.

tttzach:
@gbuenoandrade Yeah you're right, I might have twisted the definition of refactor on a lot of PRs. I'll change it!

src/main/java/com/google/sps/utility/LocationsUtility.java
```diff
@@ -63,4 +64,39 @@ private static String getLocation(Pattern pattern, String prefix, String taskNot
...
+      List<List<String>> lists, List<List<String>> result, int depth, List<String> current) {
+    if (depth == lists.size()) {
+      result.add(new ArrayList<>(current));
+      return;
+    }
+    for (int i = 0; i < lists.get(depth).size(); i++) {
+      List<String> next = new ArrayList<>();
+      next.addAll(current);
+      next.add(lists.get(depth).get(i));
+      generateCombinationsHelper(lists, result, depth + 1, next);
```
gbuenoandrade, line 87: Currently, you do O(|current|^2) work for each `current` list you end up adding to result because all the elements get copied over from current to next every time a new element gets added.
You can drop that to O(|current|) if you just append the new element to current and then delete it. I.e.,
```
current.add(lists.get(depth).get(i));
generateCombinationsHelper(lists, result, depth + 1, current);
current.remove(depth);
```
Let me know if you have any questions about this trick.

src/main/java/com/google/sps/utility/LocationsUtility.java
```diff
@@ -63,4 +64,39 @@ private static String getLocation(Pattern pattern, String prefix, String taskNot
...
+   * Generates combinations of the lists and appends them to the result parameter.
+   *
+   * @param lists A list of lists to generate combinations for.
+   * @param result A pointer to a list to contain the generated combinations.
+   * @param depth An int to represent the index of the list that is currently being traversed in the
+   *     recursive call.
+   * @param current A list containing the combination built so far at the recursive call.
+   */
+  private static void generateCombinationsHelper(
+      List<List<String>> lists, List<List<String>> result, int depth, List<String> current) {
```
gbuenoandrade, line 78: idea: `depth == current.size()` so you can drop that parameter if you want (but feel free to leave it if you think it makes things cleaner)

tttzach: I would actually prefer `depth = current.size()` and drop the parameter because then I don't have to worry about not checking the length on current in `current.remove(depth)`. Thanks!

src/main/java/com/google/sps/utility/LocationsUtility.java
tttzach: This is a smart trick, I assume it's better in terms of space complexity as well. Thanks!

src/test/java/LocationsUtilityTest.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/utility/LocationsUtility.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/utility/LocationsUtility.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/utility/LocationsUtility.java
tttzach: `RESOLVED`

src/main/java/com/google/sps/utility/GeocodingResultUtility.java
```diff
@@ -54,12 +55,9 @@
...
   public static Optional<PlaceType> convertToPlaceType(String location) {
-    for (PlaceType placeType : PlaceType.values()) {
-      if (placeType.name().equalsIgnoreCase(location.replace(" ", "_"))) {
-        return Optional.ofNullable(placeType);
-      }
-    }
-    return Optional.empty();
+    return Arrays.asList(PlaceType.values()).stream()
+        .filter(placeType -> placeType.name().equalsIgnoreCase(location.replace(" ", "_")))
+        .findFirst();
```
gbuenoandrade: Neat!
RESOLVED

src/main/java/com/google/sps/utility/LocationsUtility.java
```diff
@@ -63,4 +64,36 @@ private static String getLocation(Pattern pattern, String prefix, String taskNot
...
+      current.remove(depth);
+    }
+  }
+
+  /**
+   * Generates all combinations of the lists recursively.
+   *
+   * @param lists A list of lists to generate combinations for.
+   */
+  public static List<List<String>> generateCombinations(List<List<String>> lists) {
```
antmar, line 94: Food for thought: if you had a non-recursive implementation, then you may be able to turn this into a combination supplier for use with `Stream.generate`. This would essentially produce combinations on-demand. When producing a potentially large (like use-all-your-memory large) result, it's nice to give the caller an option to only consume some of the elements.

(No need to make this change. I am just thinking aloud.)

src/main/java/com/google/sps/utility/GeocodingResultUtility.java
tttzach: Thanks!

src/main/java/com/google/sps/utility/LocationsUtility.java
tttzach: Ohhh okay, `Stream.generate` would have been neat. Also some sort of dynamic programming would have been cool but would be a lot more complicated to implement.

tttzach:
> Can you make the PR description more specific? If I was looking at this PR in two weeks, I would have no idea what methods you are adding to Locations Utility and why.

Yeah absolutely, I'll go back and change all the PR titles and descriptions to be more descriptive. Thanks!

src/main/java/com/google/sps/utility/GeocodingResultUtility.java

src/main/java/com/google/sps/utility/LocationsUtility.java
tttzach: `RESOLVED`

PR #149, opened 2020-08-04
Implement client side for Go panel

tttzach:
Implement UI for Go panel.

Screenshot 2020-08-04 at 11 39 12 PM


src/main/webapp/index.html
```diff
@@ -166,13 +166,35 @@ <h2 class="panel-row__header u-align-with-first-panel">Build your plan:</h2>
     <div class="panel-row__content">
       <div class="panel">Assign</div>
       <div class="panel">Plan</div>
-      <div class="panel">Go
-        <div id="go"></div>
+      <div class="panel">
```
austinteshuba, line 169: Can you create a utility class `.u-border-go` and change the border of this panel to match some unique color (I'm thinking a material green for this panel - currently saved as a variable actually `--color-green`). Any color is good, but I think the black looks too plain for our colour palette.

src/main/webapp/style/base/_typography.css
```diff
@@ -49,6 +49,13 @@ body {
   width: var(--panel-icon-width);
 }
 
+.panel__optimize-button {
```
austinteshuba, line 52: So this is coming to master soon in PR #125, but there are some existing buttons you can put at the bottom of this panel. 

Alternatively, you can use this button no problem, but I would try to fix up the alignment. It seems to not be vertically centered in its container

austinteshuba: Also, it may be better to move this button to the bottom of the panel. If a tasklist name is long, it might push your button off the panel or otherwise squash it. 

You could even use the "Push to Start" button that's currently in the "Assign" panel and just create a content entry for this.

src/main/webapp/index.html
tttzach: Got it - changed to orange. Thanks!

src/main/webapp/style/base/_typography.css
tttzach: I think this is slightly better now(?)

Screenshot 2020-08-06 at 2 18 36 AM

austinteshuba: Yep it is! Small nit picks: 
1) Make the button green (there's an existing class for this, `panel__green-button`)
2) If you want, if you found a way to overwrite the `width` property of the button to be `100%` (you could make another class for this if you wanted by adopting most of the `button-1-of-2` properties and changing the width. The new class could be `button-1-of-1`)

gbuenoandrade: +1 to making the optimize button green

tttzach:
> LGTM
> 
> Though I think it would be cool if you could give the user some sort of feedback after they press "optimize." Maybe just write "..." in the meantime.

Good idea!

Before pressing optimize
Screenshot 2020-08-06 at 5 36 42 PM

After pressing optimize
Screenshot 2020-08-06 at 5 39 31 PM

src/main/webapp/style/base/_typography.css
tttzach: Thanks for the suggestions! I made a new class and put 30% instead - not a designer by any means but I just thought 100% made it look a little weird.

Screenshot 2020-08-06 at 5 38 52 PM

src/main/java/com/google/sps/servlets/GoServlet.java
```diff
@@ -148,7 +148,7 @@ public void doGet(
     List<Task> tasks;
 
     String taskLists = request.getParameter("taskLists");
-    if (taskLists == null) {
+    if (taskLists.isEmpty()) {
```
antmar, line 151: I am not sure what prompted this change, but you may need to handle both null and empty here?

For future reference, the Apache Commons package has a utility method called `StringUtils.isEmpty` that handles `null` the same as empty.

tttzach: I found that because `taskLists=` will always be in the query string, getting the `taskLists` parameter will return `""` instead of `null`. However, I changed it to `StringUtils.isEmpty` anyway just so I'll remember it exists next time. Thanks!

src/main/webapp/auth.js
```diff
@@ -79,13 +79,10 @@ function handleAuthenticationState() {
     // Populate information panels at top of dashboard
     populateGmail();
     populateTasks();
-    populateCalendar();
+    // populateCalendar();
```
antmar, line 82: Um, did you intend to comment this out? It seems unrelated to the Go panel.

tttzach: Yeah sorry about that, I wanted to screen capture the deployed site and this function writes a bunch of holidays from calendar behind the Go panel. However, we just decided it's possible to merge everything and then do our screen capture off the final product.

src/main/java/com/google/sps/servlets/GoServlet.java
tttzach: `RESOLVED`

src/main/webapp/auth.js
tttzach: `RESOLVED`

PR #150, opened 2020-08-04
Implement post request with plan-mail

AymarMuhikira:
Implement functionality to create new events to read emails when the user chooses to.
Before creating the event
Get-before
Once the event has been created
Get-after


src/main/webapp/index.html
```diff
@@ -134,7 +134,7 @@ <h2 class="panel-row__header u-align-with-first-panel">Build your plan:</h2>
         <p> Your unread emails contain <b id="word-count"></b> words
         <p>Average reply: <b id="average-reply"></b> words
         <p>Time needed: <b id="time-needed"></b> minutes
-        <p> Click to schedule
+        <p id = "message-event">
```
tttzach, line 137: nit: Remove the spaces so that it is `id="message-event"`

src/main/webapp/populateData.js
```diff
@@ -358,16 +358,30 @@ function populatePlanMail() {
...
+          messageEventContainer.innerText = 'No new events needed';
+        } else {
+          messageEventContainer.innerText = 'Click to schedule';
+          for (const index in planMailResponse.potentialEventTimes) {
+            if (typeof index === 'string') {
+              const buttonElement = document.createElement('button');
+              buttonElement.innerText =
+                planMailResponse.potentialEventTimes[index].start +
+                ' to ' +
+                planMailResponse.potentialEventTimes[index].end;
```
tttzach, line 375: Use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

src/main/java/com/google/sps/model/CalendarClient.java
```diff
@@ -58,4 +58,6 @@
    * @throws IOException thrown when an issue occurs
    */
   Date getCurrentTime() throws IOException;
+
+  void createNewEvent(Date start, Date end, String summary, String calendarId) throws IOException;
```
austinteshuba, line 62: nit: This requires Javadoc

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -123,4 +123,21 @@ public void doGet(
     }
     return events;
   }
+
+  @Override
```
austinteshuba, line 127: Some javadoc here describing what is expected in the request and what is sent back to the client would be nice.

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -123,4 +123,21 @@ public void doGet(
...
+      HttpServletRequest request, HttpServletResponse response, Credential googleCredential)
+      throws IOException {
+    assert googleCredential != null
+        : "Null credentials (i.e. unauthenticated requests) should already be handled";
+
+    CalendarClient calendarClient = calendarClientFactory.getCalendarClient(googleCredential);
+    Date start = new Date(request.getParameter("start"));
+    Date end = new Date(request.getParameter("end"));
+    String summary = "Read emails";
+    String calendarId = "primary";
```
austinteshuba, line 138: What's the meaning of "primary" here? Is this a placeholder or does this have a special meaning to Google Calendar?

src/main/java/com/google/sps/servlets/PlanMailServlet.java
```diff
@@ -104,13 +104,19 @@ public void doGet(
     FreeTimeUtility freeTimeUtility =
         new FreeTimeUtility(
             timeMin, personalBeginHour, workBeginHour, workEndHour, personalEndHour, numDays);
+    long preAssignedTime = 0;
+    String eventSummary = "Read emails";
```
austinteshuba, line 108: This method is becoming quite large, and seems to be doing multiple things. That's ok - the panel does a lot of stuff - but maybe separate some of this logic into self contained helper methods

src/main/webapp/index.html
AymarMuhikira: `Resolved`

src/main/webapp/populateData.js
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/model/CalendarClient.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/CalendarServlet.java
AymarMuhikira: Agreed, `Resolved`

src/main/java/com/google/sps/servlets/CalendarServlet.java
AymarMuhikira: It is the Id of the primary calendar. We are using the primary calendar because it is common for all users, where as other calendars might not be present for all users (eg Canadian Hollidays Calendar).

src/main/java/com/google/sps/servlets/PlanMailServlet.java
AymarMuhikira: The `for-loop` is doing a lot, I agree. But it would be difficult to implement a helper function for this because we are essentially getting two separate variables out of it.

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -123,4 +123,29 @@ public void doGet(
     }
     return events;
   }
+
+  /**
+   * Creates a new event when POST method called. Use the time boundaries given in the request for
+   * creating the event in the primary calendar. The summary for this event is Read emails.
+   *
+   * @param request Http request from the client. Body contains start and end parameters
+   * @param response 403 if user is not authenticated, or string saying the event has been created
```
antmar, line 132: Documentation: you are not actually returning 403 in the response here, and this method won't be called if a 403 is returned by the base class.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -103,13 +103,19 @@ public void doGet(
     FreeTimeUtility freeTimeUtility =
         new FreeTimeUtility(
             timeMin, personalBeginHour, workBeginHour, workEndHour, personalEndHour, numDays);
+    long preAssignedTime = 0;
+    String eventSummary = "Read emails";
```
antmar, line 107: How can you make sure the event summary that you are looking for is consistent with the event summary that you use when creating new events?

src/main/java/com/google/sps/servlets/CalendarServlet.java
antmar: This should be a class constant.

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -123,4 +123,29 @@ public void doGet(
...
+  public void doPost(
+      HttpServletRequest request, HttpServletResponse response, Credential googleCredential)
+      throws IOException {
+    assert googleCredential != null
+        : "Null credentials (i.e. unauthenticated requests) should already be handled";
+
+    CalendarClient calendarClient = calendarClientFactory.getCalendarClient(googleCredential);
+    Date start = new Date(request.getParameter("start"));
+    Date end = new Date(request.getParameter("end"));
+    String summary = "Read emails";
```
antmar, line 145: It's weird that the generic `CalendarServlet` always creates events with the summary "Read emails". I think that the client should supply this value. Of course you also need to make sure it's consistent with the event summary that `FreeTimeServlet` expects.

src/main/webapp/populateData.js
```diff
@@ -385,16 +385,30 @@ function populatePlanMail() {
...
+                `${planMailResponse.potentialEventTimes[index].start} to \
+                ${planMailResponse.potentialEventTimes[index].end}`;
+              buttonElement.setAttribute('start',
+                  planMailResponse.potentialEventTimes[index].start);
+              buttonElement.setAttribute('end',
+                  planMailResponse.potentialEventTimes[index].end);
+              buttonElement.addEventListener('click', () => {
+                createEvent(buttonElement.getAttribute('start'),
+                    buttonElement.getAttribute('end'));
+              });
```
antmar, line 409: You could simplify this by just capturing the start/end values in the event listener callback, but what you have is technically correct (and maybe useful in the future), so up to you.

src/main/java/com/google/sps/servlets/CalendarServlet.java
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: The summary for the events we are creating are the same as this one. For now this is the check we are using. We assume that the user will not create events with the same summary if they are not related to reading emails. I will add a comment before the loop.

src/main/java/com/google/sps/servlets/CalendarServlet.java
AymarMuhikira: I agree that this would give the user more control, but I don't know that we could capture all the events created in this process in `PlanMailServlet` (i.e. what happens if the user changes the summary, but we still need to account for events created with the previous summary?)

src/main/webapp/populateData.js
AymarMuhikira: I think I tried this in the walkthrough projects. If I supply directly the values in the event listener, it will only capture the last values of the `const`. For safety purposes, I first linked the parameters as attributes of the button, that way the values are not impacted by the `for loop`.

src/main/java/com/google/sps/servlets/CalendarServlet.java
antmar: I am not saying the user should specify the event summary. I am just saying it should come from the JS client. You can hardcode on the client (as a temporary solution) or have the PlanMailServlet tell the client what summary to use.

src/main/java/com/google/sps/servlets/PlanMailServlet.java
```diff
@@ -63,6 +63,7 @@
   private static final int workEndHour = 18;
   private static final int personalEndHour = 23;
   private static final int numDays = 5;
+  private static final String eventSummary = "Read emails";
```
gbuenoandrade, line 66: nit: please use CONSTANT_CASE

src/main/webapp/populateData.js
antmar: Oh yes, I see what the problem is. I think there is a way to work around it with temporary variables, but you can just leave it as is.

src/main/webapp/populateData.js
```diff
@@ -124,7 +124,7 @@ function populateCalendar() {
         const panelContent = document.querySelector('#panel-content');
         panelContent.innerHTML = '';
         for (const day in hoursJson.workTimeFree) {
-          if (typeof day == 'string') {
+          if (typeof day === 'string') {
```
gbuenoandrade, line 127: What else could `day` be? That looks a bit weird, but I'm okay with it if it's the simplest way.

src/main/webapp/populateData.js
```diff
@@ -337,6 +337,22 @@ function populatePlanMail() {
...
+ * @param {string} eventStart the start time of the event to create
+ * @param {string} eventEnd the end time of the event to create
+ */
+function createEvent(eventStart, eventEnd) {
+  const params = new URLSearchParams();
+  params.append('start', eventStart);
+  params.append('end', eventEnd);
+  fetch('/calendar', {method: 'POST', body: params});
+  populateCalendar();
+  populatePlanMail();
```
gbuenoandrade, line 353: would you still want to populate those panels if the fetch above failed?

src/main/webapp/populateData.js
antmar: I thought there was a different check that you were doing, as in the other PR?

src/main/java/com/google/sps/servlets/PlanMailServlet.java
AymarMuhikira: `Resolved`

src/main/webapp/populateData.js
AymarMuhikira: The code here got reverted. I don't know what hapened exactly, but I re-implemented the check

src/main/webapp/populateData.js
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/CalendarServlet.java
AymarMuhikira: I am now passing the event summary from `populatePlanMail`

src/main/webapp/populateData.js
```diff
@@ -299,7 +300,7 @@ function populateGo() {
  */
 function populatePlanMail() {
   const planContainer = document.querySelector('#plan');
-  fetch('/plan-mail')
+  fetch('/plan-mail?summary=Read emails')
```
antmar, line 303: Including the summary for `PlanMailServlet` to use is a good idea. Can you do the same in the POST to `CalendarServlet`?

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -44,6 +44,8 @@
   private static final int workEndHour = 18;
   private static final int personalEndHour = 23;
   private static final int numDays = 5;
+  private static final String summary = "Read emails";
+  private static final String calendarId = "primary";
```
antmar, line 48: Nit: PRIMARY_CALENDAR_ID

antmar, line 48: Also fix capitalization, for all constants in this class.

src/main/webapp/populateData.js
AymarMuhikira: `Resolved`

src/main/java/com/google/sps/servlets/CalendarServlet.java
AymarMuhikira: Changed it to a variable that gets the parameter from js.

AymarMuhikira: `Resolved`

PR #153, opened 2020-08-05
Return sender with actionable messages

austinteshuba:
The sender of an email will now be sent with the ActionableMessage response. This allows me to display the sender alongside the subject line on the client. Changes made due to suggestions in PR #137


src/main/java/com/google/sps/model/ActionableMessage.java
```diff
@@ -44,9 +47,13 @@ public String getSubject() {
...
   }
 
+  public String getSender() {
+    return sender;
+  }
+
   @Override
   public int hashCode() {
-    return new HashCodeBuilder(7, 17).append(id).append(subject).toHashCode();
+    return new HashCodeBuilder(7, 17).append(id).append(subject).append(sender).toHashCode();
```
gbuenoandrade, line 56: nit: why not the default `HashCodeBuilder` constructor?

src/main/java/com/google/sps/utility/GmailUtility.java
```diff
@@ -52,4 +52,21 @@ public static MessagePartHeader extractHeader(Message message, String headerName
...
+   *
+   * <p>"From" header values have two possible formats: "<sampleemail@sample.com>" (if name is not
+   * available) OR "Sample Name <sampleemail@sample.com>" If a name is available, this is extracted.
+   * Otherwise, the email is extracted
+   *
+   * @param fromHeaderValue the value of a "From" header from which a contact name / email should be
+   *     extracted
+   * @return A contact name if available, or an email if it is not
+   */
+  public static String parseNameInFromHeader(String fromHeaderValue) {
```
antmar, line 67: Consider using a regular expression instead - it would be more readable than parsing by hand.

src/main/java/com/google/sps/model/ActionableMessage.java
antmar: +1 to using the default constructor

austinteshuba: Agreed - thanks!

src/main/java/com/google/sps/utility/GmailUtility.java
austinteshuba: Done!

src/main/java/com/google/sps/utility/GmailUtility.java
```diff
@@ -63,10 +65,21 @@ public static MessagePartHeader extractHeader(Message message, String headerName
    * @param fromHeaderValue the value of a "From" header from which a contact name / email should be
    *     extracted
    * @return A contact name if available, or an email if it is not
+   * @throws IllegalArgumentException If value could not be parsed from header
    */
   public static String parseNameInFromHeader(String fromHeaderValue) {
-    return fromHeaderValue.charAt(0) == '<'
-        ? fromHeaderValue.substring(1, fromHeaderValue.length() - 1)
-        : fromHeaderValue.split("<")[0].trim();
+    Pattern withNamePattern = Pattern.compile("(.*) <.*>");
```
antmar, line 71: Make the Pattern a static variable. The `Pattern.compile` is a relatively expensive operation. If you have a constant pattern, it's best to only compile it once.

src/main/java/com/google/sps/utility/GmailUtility.java
```diff
@@ -63,10 +65,21 @@ public static MessagePartHeader extractHeader(Message message, String headerName
...
+    Matcher withNameMatcher = withNamePattern.matcher(fromHeaderValue);
+    if (withNameMatcher.find()) {
+      return withNameMatcher.group(1);
+    }
+
+    Pattern withoutNamePattern = Pattern.compile("<(.*)>");
```
antmar, line 77: I think you could do this with a single pattern: `(.*)?<(.*)>`
Use `groupCount` to check if there are 1 or 2 groups in the match. Don't forget to trim the name afterwards. (You could make it more complex to avoid the need to trim, but trimming is probably a good idea anyway.)

src/main/java/com/google/sps/utility/GmailUtility.java
austinteshuba: Okay interesting! Good suggestion - thanks

src/main/java/com/google/sps/utility/GmailUtility.java
austinteshuba: Good suggestion - I can't do this check directly with `groupCount`, as group count simply returns the number of capturing groups in the pattern, which is always 2. However, I was able to do make this work by extracting the name group and email group, and returning the appropriate one depending on if the extracted name is empty.

PR #154, opened 2020-08-06
Implement tests for the FreeTime servlet

AymarMuhikira:
-Refactor getWordCount method to be able to mock it
-Implement tests for the servlet

### DIFF
--- a/src/main/java/com/google/sps/model/GmailClient.java
+++ b/src/main/java/com/google/sps/model/GmailClient.java
@@ -55,6 +55,16 @@ public interface GmailClient {
       throws IOException;
 
   /**
+   * Gets the word count in the unread emails from the past days. The period length to go
+   * through the emails is specified in the parameter passed in the method
+   *
+   * @param numberDaysUnread the number of days to look through the emails
+   * @return an integer representing the word count in the unread emails
+   * @throws IOException if an issue occurs with the Gmail service
+   */
+  int getWordCount(int numberDaysUnread) throws IOException;
+
+  /**
    * Encapsulates possible values for the "format" query parameter in the Gmail GET message method
    */
   enum MessageFormat {
--- a/src/main/java/com/google/sps/model/GmailClient.java
+++ b/src/main/java/com/google/sps/model/GmailClient.java
@@ -55,8 +55,8 @@ public interface GmailClient {
       throws IOException;
 
   /**
-   * Gets the word count in the unread emails from the past days. The period length to go
-   * through the emails is specified in the parameter passed in the method
+   * Gets the word count in the unread emails from the past days. The period length to go through
+   * the emails is specified in the parameter passed in the method
    *
    * @param numberDaysUnread the number of days to look through the emails
    * @return an integer representing the word count in the unread emails
--- a/src/main/java/com/google/sps/model/GmailClientImpl.java
+++ b/src/main/java/com/google/sps/model/GmailClientImpl.java
@@ -23,9 +23,14 @@ import com.google.api.services.gmail.Gmail;
 import com.google.api.services.gmail.model.Message;
 import com.google.sps.utility.ServletUtility;
 import java.io.IOException;
+import javax.mail.MessagingException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
+import com.google.api.services.gmail.model.MessagePart;
+import com.google.common.io.BaseEncoding;
+import com.google.api.client.util.StringUtils;
+import java.util.StringTokenizer;
 
 /** Handles GET requests from Gmail API */
 public class GmailClientImpl implements GmailClient {
@@ -123,6 +128,48 @@ public class GmailClientImpl implements GmailClient {
               }
             })
         .collect(Collectors.toList());
+  }
+
+  @Override
+  public int getWordCount(int numberDaysUnread) {
+    GmailClient.MessageFormat messageFormat = GmailClient.MessageFormat.FULL;
+    int wordCount = 0;
+    List<Message> unreadMessages = new ArrayList<>();
+    try {
+      unreadMessages = getUnreadEmailsFromNDays(messageFormat, numberDaysUnread);
+    } catch (IOException e) {
+      System.out.println(e);
+    }
+    for (Message message : unreadMessages) {
+      try {
+        wordCount += getMessageSize(message);
+      } catch (MessagingException | IOException e) {
+        System.out.println(e);
+      }
+    }
+    return wordCount;
+  }
+
+  /**
+   * Private method to get the individual word count of a single email
+   */
+  private int getMessageSize(Message message) throws MessagingException, IOException {
+    List<MessagePart> messageParts = message.getPayload().getParts();
+    List<MessagePart> messageBody =
+        messageParts.stream()
+            .filter((messagePart) -> messagePart.getMimeType().equals("text/plain"))
+            .collect(Collectors.toList());
+    int size = 0;
+    if (messageBody.isEmpty()) {
+      return size;
+    }
+    for (MessagePart part : messageBody) {
+      byte[] messageBytes = BaseEncoding.base64Url().decode(part.getBody().getData());
+      String messageString = StringUtils.newStringUtf8(messageBytes);
+      StringTokenizer tokens = new StringTokenizer(messageString);
+      size += tokens.countTokens();
+    }
+    return size;
   }
 
   /**
--- a/src/main/java/com/google/sps/model/GmailClientImpl.java
+++ b/src/main/java/com/google/sps/model/GmailClientImpl.java
@@ -19,18 +19,18 @@ import com.google.api.client.extensions.appengine.http.UrlFetchTransport;
 import com.google.api.client.http.HttpTransport;
 import com.google.api.client.json.JsonFactory;
 import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.client.util.StringUtils;
 import com.google.api.services.gmail.Gmail;
 import com.google.api.services.gmail.model.Message;
-import com.google.sps.utility.ServletUtility;
-import java.io.IOException;
-import javax.mail.MessagingException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.stream.Collectors;
 import com.google.api.services.gmail.model.MessagePart;
 import com.google.common.io.BaseEncoding;
-import com.google.api.client.util.StringUtils;
+import com.google.sps.utility.ServletUtility;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.StringTokenizer;
+import java.util.stream.Collectors;
+import javax.mail.MessagingException;
 
 /** Handles GET requests from Gmail API */
 public class GmailClientImpl implements GmailClient {
@@ -150,9 +150,7 @@ public class GmailClientImpl implements GmailClient {
     return wordCount;
   }
 
-  /**
-   * Private method to get the individual word count of a single email
-   */
+  /** Private method to get the individual word count of a single email */
   private int getMessageSize(Message message) throws MessagingException, IOException {
     List<MessagePart> messageParts = message.getPayload().getParts();
     List<MessagePart> messageBody =
--- a/src/main/java/com/google/sps/servlets/FreeTimeServlet.java
+++ b/src/main/java/com/google/sps/servlets/FreeTimeServlet.java
@@ -16,12 +16,9 @@ package com.google.sps.servlets;
 
 import com.google.api.client.auth.oauth2.Credential;
 import com.google.api.client.util.DateTime;
-import com.google.api.client.util.StringUtils;
 import com.google.api.services.calendar.model.CalendarListEntry;
 import com.google.api.services.calendar.model.Event;
 import com.google.api.services.gmail.model.Message;
-import com.google.api.services.gmail.model.MessagePart;
-import com.google.common.io.BaseEncoding;
 import com.google.sps.data.PlanMailResponse;
 import com.google.sps.model.AuthenticatedHttpServlet;
 import com.google.sps.model.AuthenticationVerifier;
@@ -39,10 +36,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
-import java.util.StringTokenizer;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
-import javax.mail.MessagingException;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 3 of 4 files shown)


src/main/java/com/google/sps/model/GmailClient.java
```diff
@@ -54,6 +54,16 @@
...
 
+  /**
+   * Gets the word count in the unread emails from the past days. The period length to go through
+   * the emails is specified in the parameter passed in the method
+   *
+   * @param numberDaysUnread the number of days to look through the emails
+   * @return an integer representing the word count in the unread emails
+   * @throws IOException if an issue occurs with the Gmail service
+   */
+  int getWordCount(int numberDaysUnread) throws IOException;
```
austinteshuba, line 65: I would abstract this into a different helper than this one - this is meant to be for GET requests from Gmail, while this deals with the processing of information. Consider creating a `PlanMailResponseHelper` class, similar to what was done for Gmail. Or other names if you have any ideas

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
```diff
@@ -125,9 +119,11 @@ public void doGet(
       freeTimeUtility.addEvent(eventStart, eventEnd);
     }
 
-    int wordCount = getWordCount(googleCredential);
+    int numberDaysUnread = 7;
+    int wordCount = gmailClient.getWordCount(numberDaysUnread);
     int averageReadingSpeed = 50;
-    int minutesToRead = (int) Math.ceil((double) wordCount / averageReadingSpeed);
+    int minutesToRead;
+    minutesToRead = wordCount == 0 ? 0 : (int) Math.ceil((double) wordCount / averageReadingSpeed);
```
austinteshuba, line 126: nit: simplify to `(int) Math.ceil((double) wordCount / averageReadingSpeed);`

src/test/java/FreeTimeServletTest.java
```diff
@@ -0,0 +1,211 @@
...
+
+  @Test
+  public void noPotentialEvent() throws Exception {
+    // Test case where there are no unread emails
+    // We should get no proposed potential event times
+    Mockito.when(calendarClient.getCalendarList()).thenReturn(ONE_CALENDAR);
+    Mockito.when(calendarClient.getUpcomingEvents(PRIMARY, CURRENT_TIME, END_TIME))
+        .thenReturn(NO_EVENT);
+    Mockito.when(calendarClient.getCurrentTime()).thenReturn(CURRENT_TIME);
+    Mockito.when(gmailClient.getWordCount(7)).thenReturn(0);
```
austinteshuba, line 122: What is the meaning of this `7`? I would consider changing this to a constant as this seems non-obvious

src/test/java/FreeTimeServletTest.java
```diff
@@ -0,0 +1,211 @@
...
+
+  @Test
+  public void onePotentialEvent() throws Exception {
+    // Test case where there are unread emails that will take 10 minutes to read
+    // We should get 1 proposed event that takes 10 minutes
+    Mockito.when(calendarClient.getCalendarList()).thenReturn(ONE_CALENDAR);
+    Mockito.when(calendarClient.getUpcomingEvents(PRIMARY, CURRENT_TIME, END_TIME))
+        .thenReturn(EVENT_ONE);
+    Mockito.when(calendarClient.getCurrentTime()).thenReturn(CURRENT_TIME);
+    Mockito.when(gmailClient.getWordCount(7)).thenReturn(500);
```
austinteshuba, line 138: Ditto to above

src/test/java/FreeTimeServletTest.java
```diff
@@ -0,0 +1,211 @@
...
+    PlanMailResponse actual = getServletResponse();
+    List<Date> startHour = Arrays.asList(new Date(2020 - OFFSET_YEAR, 4, 19, 10, 0));
+    List<Date> endHour = Arrays.asList(new Date(2020 - OFFSET_YEAR, 4, 19, 10, 10));
+    Assert.assertEquals(500, actual.getWordCount());
+    Assert.assertEquals(50, actual.getAverageReadingSpeed());
+    Assert.assertEquals(10, actual.getMinutesToRead());
+    Assert.assertEquals(startHour.size(), actual.getPotentialEventTimes().size());
+    for (int index = 0; index < startHour.size(); index++) {
+      Assert.assertTrue(
```
austinteshuba, line 147: nit: use `Assert.assertEquals`

src/test/java/FreeTimeServletTest.java
```diff
@@ -0,0 +1,211 @@
...
+            new Date(2020 - OFFSET_YEAR, 4, 19, 10, 30));
+    List<Date> endHour =
+        Arrays.asList(
+            new Date(2020 - OFFSET_YEAR, 4, 19, 10, 5),
+            new Date(2020 - OFFSET_YEAR, 4, 19, 10, 35));
+    Assert.assertEquals(500, actual.getWordCount());
+    Assert.assertEquals(50, actual.getAverageReadingSpeed());
+    Assert.assertEquals(10, actual.getMinutesToRead());
+    Assert.assertEquals(startHour.size(), actual.getPotentialEventTimes().size());
+    for (int index = 0; index < startHour.size(); index++) {
```
austinteshuba, line 196: Instead of this loop, consider overriding the `public int hashCode()` and `public boolean equals()` methods to determine equality. Take a look at `ActionableMessage`, as I did this there with some helpful Apache tools

You can then just use `Assert.assertEquals(expectedPlanMailResponse, actualPlanMailResponse)`, which is more readable. 

Don't spend too much time on this unless if @gbuenoandrade or @antmar thinks this is a high-value thing to do, but this might help clean things up and avoid using a loop for assertions.

src/main/java/com/google/sps/model/GmailClient.java
antmar: Drive-by comment - I agree about putting these methods in a separate utility. 

Also, it's not immediately obvious that the implementation of this method is tested. A separate utility with separate tests would make that obvious.

src/main/java/com/google/sps/model/GmailClientImpl.java
```diff
@@ -125,6 +130,46 @@ public Message getUserMessageWithMetadataHeaders(String messageId, List<String>
...
+    } catch (IOException e) {
+      System.out.println(e);
+    }
+    for (Message message : unreadMessages) {
+      try {
+        wordCount += getMessageSize(message);
+      } catch (MessagingException | IOException e) {
+        System.out.println(e);
+      }
+    }
```
tttzach, line 149: I think something like this is a little bit more cleaner
```
try {
  getUnreadEmailsFromNDays(messageFormat, numberDaysUnread)
    .stream()
    .map(message -> wordCount += getMessageSize(message));
} catch (MessagingException | IOException e) {
  System.out.println(e);
}
```

src/test/java/FreeTimeServletTest.java
```diff
@@ -0,0 +1,211 @@
...
+        .thenReturn(EVENT_ONE);
+    Mockito.when(calendarClient.getUpcomingEvents(SECONDARY, CURRENT_TIME, END_TIME))
+        .thenReturn(EVENT_TWO);
+    Mockito.when(calendarClient.getCurrentTime()).thenReturn(CURRENT_TIME);
+    Mockito.when(gmailClient.getWordCount(7)).thenReturn(500);
+    PlanMailResponse actual = getServletResponse();
+    Assert.assertEquals(500, actual.getWordCount());
+    Assert.assertEquals(50, actual.getAverageReadingSpeed());
+    Assert.assertEquals(10, actual.getMinutesToRead());
+    Assert.assertTrue(actual.getPotentialEventTimes().isEmpty());
```
tttzach, line 170: If you override `equals` in `PlanMailResponse`, you can do `Assert.assertEquals(response1, response2)` but it's up to you, it's the last day haha

src/main/java/com/google/sps/model/GmailClient.java
AymarMuhikira: I agree, I refactored my implementation to put this in `PlanMailResponseHelper`. As for testing the method itself, that requires creating Message objects to pass in, and I am having run-time errors when I do that. I am tabling it so that I can focus on the rest of the review for now.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
AymarMuhikira: Agreed. I was trying to account for when averageReadingSpeed would be 0, but that should never be the case to begin with. This was redundant. `Resolved`

src/test/java/FreeTimeServletTest.java
AymarMuhikira: `Resolved`

src/test/java/FreeTimeServletTest.java
AymarMuhikira: `Resolved`

src/test/java/FreeTimeServletTest.java
AymarMuhikira: `Resolved`

src/test/java/FreeTimeServletTest.java
AymarMuhikira: I overrode the equal method in `PlanMailResponse`, `Resolved`

src/test/java/FreeTimeServletTest.java
AymarMuhikira: I agree, it's much cleaner. `Resolved`

src/main/java/com/google/sps/model/GmailClientImpl.java
AymarMuhikira: I have now separated the logic to get the List of messages and to do the word count in two different classes. I will try to implement this for earlier PRs.

src/main/java/com/google/sps/servlets/FreeTimeServlet.java
gbuenoandrade: Out of curiosity, to avoid manipulating floating point numbers, you can use the follow identity:
`ceil(a/b) = floor((a+b-1)/b)` (and the floor is what an integer division will give you by default in most languages.)

src/main/java/com/google/sps/model/GmailClientImpl.java
tttzach: Thanks! `RESOLVED`

src/test/java/FreeTimeServletTest.java
tttzach: Thanks! `RESOLVED`

src/main/java/com/google/sps/model/GmailClient.java
austinteshuba: Sure 

RESOLVED

src/test/java/FreeTimeServletTest.java
austinteshuba: Cool! Thanks

src/test/java/PlanMailServletTest.java
```diff
@@ -0,0 +1,221 @@
...
+        .thenReturn(MESSAGES);
+    Mockito.when(planMailResponseHelper.getWordCount(Mockito.any())).thenReturn(TEST_WORD_COUNT);
+    PlanMailResponse actual = getServletResponse();
+    List<Date> startHour = Arrays.asList(new Date(2020 - OFFSET_YEAR, 4, 19, 10, 0));
+    List<Date> endHour = Arrays.asList(new Date(2020 - OFFSET_YEAR, 4, 19, 10, 10));
+    List<DateInterval> meetingTimes = new ArrayList<>();
+    for (int index = 0; index < startHour.size(); index++) {
+      meetingTimes.add(new DateInterval(startHour.get(index), endHour.get(index)));
+    }
+    PlanMailResponse expected = new PlanMailResponse(500, 50, 10, meetingTimes);
```
antmar, line 161: Use TEST_WORD_COUNT instead of 500.

src/test/java/PlanMailServletTest.java
```diff
@@ -0,0 +1,221 @@
...
+    Mockito.when(calendarClient.getUpcomingEvents(PRIMARY, CURRENT_TIME, END_TIME))
+        .thenReturn(EVENT_ONE);
+    Mockito.when(calendarClient.getUpcomingEvents(SECONDARY, CURRENT_TIME, END_TIME))
+        .thenReturn(EVENT_TWO);
+    Mockito.when(calendarClient.getCurrentTime()).thenReturn(CURRENT_TIME);
+    Mockito.when(gmailClient.getUnreadEmailsFromNDays(MESSAGE_FORMAT, NUMBER_DAYS_UNREAD))
+        .thenReturn(MESSAGES);
+    Mockito.when(planMailResponseHelper.getWordCount(Mockito.any())).thenReturn(TEST_WORD_COUNT);
+    PlanMailResponse actual = getServletResponse();
+    PlanMailResponse expected = new PlanMailResponse(500, 50, 10, new ArrayList<>());
```
antmar, line 180: Use TEST_WORD_COUNT instead of 500.

src/test/java/PlanMailServletTest.java
```diff
@@ -0,0 +1,221 @@
...
+    PlanMailResponse actual = getServletResponse();
+    List<Date> startHour =
+        Arrays.asList(
+            new Date(2020 - OFFSET_YEAR, 4, 19, 10, 0),
+            new Date(2020 - OFFSET_YEAR, 4, 19, 10, 30));
+    List<Date> endHour =
+        Arrays.asList(
+            new Date(2020 - OFFSET_YEAR, 4, 19, 10, 5),
+            new Date(2020 - OFFSET_YEAR, 4, 19, 10, 35));
+    Assert.assertEquals(500, actual.getWordCount());
```
antmar, line 205: This assert seems redundant, because you are testing the entire content of the actual response later.

src/test/java/PlanMailServletTest.java
```diff
@@ -0,0 +1,221 @@
...
+    List<DateInterval> meetingTimes = new ArrayList<>();
+    for (int index = 0; index < startHour.size(); index++) {
+      meetingTimes.add(new DateInterval(startHour.get(index), endHour.get(index)));
+    }
+    PlanMailResponse expected = new PlanMailResponse(500, 50, 10, meetingTimes);
```
antmar, line 210: Use TEST_WORD_COUNT instead of 500.

src/main/java/com/google/sps/data/PlanMailResponse.java
```diff
@@ -52,4 +52,48 @@ public int getMinutesToRead() {
   public List<DateInterval> getPotentialEventTimes() {
     return potentialEventTimes;
   }
+
+  @Override
+  public boolean equals(Object o) {
```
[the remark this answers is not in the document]
austinteshuba: +1 to antmar - it's super helpful. 

While you do this, please also override the `hashcode` function - there's a builder for that too and it is required to fulfill the "contract" in Java which states that equal objects will have the same Hashcode

